PR #6, opened 2020-04-10
Static code analysis

tankefugl:
Added static code analysis using cppcheck. 
Added vscode remote contaner development support.


README.md
```diff
@@ -83,6 +83,16 @@ Short vocabulary
...
 ==================
 
+Quick start with VS Code and Remote Container
+---------------------------------------------
+
+1. Download and install [Visual Studio Code](https://code.visualstudio.com/).
+2. Install the extension __Remote Development__.
+3. Clone _dlite_ and initialize git modules: `git submodule update --init`.
+4. Open the _dlite_ folder with VS Code.
+5. When presented with a notification, select __Reopen in container__. This will build the container and restart VS Code in it. This may take some time the first time as the Docker image must be built.
```
jesper-friis, line 93: Should there be a point about installing Docker?

I do not get a notification after opening _dlite_ in VS Code. It might be because that I am using [Fedora 31, which enables cgroup v2](https://www.redhat.com/sysadmin/fedora-31-control-group-v2) by default. To use Docker, it seems that I have to [revert back to cgroup v1](https://linuxconfig.org/how-to-install-docker-on-fedora-31).

.github/workflows/ci_tests.yml
```diff
@@ -1,49 +0,0 @@
...
-        cd -
-    - name: make
-      run: |
-        cd build
-        make
-        cd -
-    - name: make test
-      run: |
-        cd build
-        make test
```
jesper-friis, line 48: What is the reason for removing this workflow? Isn't is a very useful complement to the static code analyser to build and run the test suite for each push?

tankefugl: Yes, but it is identical to the workflow in ci.yml, which already does testing and all that jazz. So, I thought I'd remove one of them assuming that having them both saves no real purpose 🥦 !

README.md
tankefugl: Ok, I might revise this description a bit. You should not need the notificaiton to open it.

tankefugl: Also, I will take a look at the false positives. It is best if we can avoid adding suppressions, but getting checking in has value even with supression of some issues.

jesper-friis: I have reverted to cgroup v1 and can now run docker again! I still don't get any __Reopen in container__ option in VS Code, but the docker tab shows a list of containers and images. Is there a menu option?

tankefugl:
Continuing this after the latest fixes from Jesper has been merged from a sibling branch.

jesper-friis:
Link plugins using shared libraries as described in issue https://github.com/SINTEF/dlite/issues/9. 
This works fine on Fedora 31 and in the docker container (after passing the RTLD_GLOBAL flag to dlopen().  We leave it to issue #9 to fix it on Windows.|

jesper-friis:
Why is there a third stale workflow? Is it because we require all tests to pass, but  .github/workflows/dockerimage.yml is configured to only run when we push to master or add a version tag?

tankefugl:
Yes, that is most likely it.

I suggest that we change the third check and build the docker image (with tests!) on every change, just don't push it unless it is a tagged release. The job steps have conditions for this. I can take a look at that and do it in this PR (later today?) unless you beat me to it =)

jesper-friis:
We can do that, but I guess that the docker test fail before we have merged to master, unless we replace 

    RUN git clone https://github.com/SINTEF/dlite.git

with

    COPY --chown=user:user . dlite

in the Dockerfile. Is there a preferred way? The latter has the benefit that it is possible to debug locally before pushing to github (faster and creates less noise).

tankefugl:
He latter way is preferred, it also means that the Docker container can be decoupled from the exact git branch/commit, and changing the code will invalidate the container (COPY vs. RUN)

.github/workflows/dockerimage.yml
```diff
@@ -28,17 +28,32 @@ jobs:
...
-#         # Publish
-#         echo "Publish docker image: docker.pkg.github.com/sintef/dlite/$IMAGE_NAME:$(<VERSION)"
-#         docker push docker.pkg.github.com/sintef/dlite/$IMAGE_NAME:$(<VERSION)
+    - name: Create github docker package
+      uses: whoan/docker-build-with-cache-action@v4
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}  # token provided by actions
+      with:
+        push_image_and_stages: false
+        username: jesper-friis
```
jesper-friis, line 34: Is possible to not hard code my username here or use sintef as username?
What happens if people other than me are pushing, will this workflow then fail or are the credential provided in the token used to publish using my username?

CMakeLists.txt
```diff
@@ -6,7 +6,7 @@ cmake_minimum_required(VERSION 3.11)
 include(CPack)
 
 project(dlite
-  VERSION 0.2.2-alpha
+  VERSION 0.2.2.1
```
jesper-friis, line 9: Doesn't semantic versioning require hypen instead of dot between patch version number and pre-release version? See point 9 on https://semver.org/

.github/workflows/dockerimage.yml
jesper-friis: I think we can ignore my comment above, since there must be a real user behind publishing a docker container. The workflow passes also when you are triggering it. Wonderful work Petter!

PR #7, opened 2020-04-10
Fixed issues found by cppcheck and added it to standard test suite

jesper-friis:
- Fixed issues found by cppcheck.
- Suppressed some false positives using `--inline-suppr` and inline comments in the code
- Added cppcheck to standard ctest test suite

### DIFF
--- a/.github/workflows/static_analysis.yml
+++ b/.github/workflows/static_analysis.yml
@@ -21,4 +21,4 @@ jobs:
 
     - name: Run static code analysis
       run: |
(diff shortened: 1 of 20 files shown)


tankefugl:
When testing the changes I noticed a potential security issue (GCC warned me). I turned warnings into errors and attempted the fix. @jesper-friis please take a look at the change for me.

tankefugl:
For some reason, the CI now fails with new warnings! Not sure why.

jesper-friis:
Thanks for the fixes. I found another error by compiling with with -Werror.

PR #32, opened 2020-08-20
Added some general functions for copying and casting arrays of n dimensions

jesper-friis:
Needed for generic Fortran bindings


jesper-friis:
Replaced by PR #33

PR #38, opened 2020-09-07
Fortran bindings

terence-coudert:
There is one issue for 2D arrays.


bindings/fortran/tests/Person.f90
```diff
@@ -0,0 +1,250 @@
+!
```
jesper-friis, line 1: This file can be removed from the source directory, since it now is generated

bindings/fortran/tests/test_person.f90
```diff
@@ -0,0 +1,78 @@
...
+  scan%points(1,1) = 1
+  scan%points(1,2) = 2
+  scan%points(1,3) = 3
+  print *, "Array, shape=(", size(scan%points, 1), ",", size(scan%points, 2), ")"
+  do i = 1, size(scan%points, 1)
+    print *, scan%points(i, :)
+  end do
+  status = scan%writeToURL("json://scans2.json")
+
+  !person%destroy()
```
jesper-friis, line 75: Running the tests with valgrind shows some memory leaks. It is possible to add these destroy() methods?

tools/templates/fortran-module.txt
```diff
@@ -0,0 +1,276 @@
...
+{@40}:: {prop.name}{@if:{prop.ndims}!0}({prop.dims::{,}\.}){@endif}\
+ ! {prop.descr} {@if:{prop.ndims}!0} [{prop.dims:{dim.name}{,}\.}]{@endif}\n}
+    contains
+      procedure :: check => check
+      procedure :: writeToURL => writeToURL
+      procedure :: writeToSource => writeToSource
+      procedure :: writeToStorage => writeToStorage
+      procedure :: readFromInstance => readFromInstance
+      procedure :: writeToInstance => writeToInstance
+      !procedure :: destroy => destroy
```
jesper-friis, line 52: Can we also generate a destroy() method?

jesper-friis:
Looks very good. Is it easy to add the destroy method mentioned in the comments above?

PR #46, opened 2020-12-06
Added Finddlite.cmake file to cmake/ directory.

jesper-friis:
Added a short note about environment variables in the README file.


terence-coudert (changes requested):
missing the variable dlite_TEMPLATES that is used by the cmake macro dlite_codegen

jesper-friis:
Running

    find -type f | grep -v build xargs grep dlite_TEMPLATES

in the root directory doesn't shown any matches. Do you DLITE_TEMPLATE_DIRS?

PR #47, opened 2020-12-07
Fixed failure in test: test_postgresql_storage2

jesper-friis:


terence-coudert:
How to enable the tests test_postgresql_storage and test_postgresql_storage2 ?

jesper-friis:
You need to have a postgresql server running. Then go to storages/python/tests and add a pgconf file as described in CMakeLists.txt

PR #49, opened 2020-12-08
Fixed test_metamodel

jesper-friis:


terence-coudert:
test_metamodel in c and test_animal in fortran passed

jesper-friis:
Great, I think we can merge then

PR #88, opened 2021-05-05
updated with files

jesper-friis:
Made this a PR so that it is checked via github actions


jesper-friis:
Running pip as root seems to be discouraged. Is it ok in this case as we already are in a controlled environment?

pip install gives the following warning:
```
WARNING: Running pip as root will break packages and permissions. You should install packages reliably by using venv: https://pip.pypa.io/warnings/venv
```

jesper-friis:
I had to enable the bionic repo - now it works

PR #89, opened 2021-05-05
Added dockerfile for cross-compilation for Windows

jesper-friis:
We use a Fedora image for this, since Fedora has better support for the mingw toolsuite than Ubuntu.

Also fixed a lot of portability issues.


jesper-friis:
This seems to work well, so I am merging this to master. now.

PR #111, opened 2021-06-25
Msvc native build; python setup.py install

pscff:
* **Changes to CMakeList.txt for building with "Visual Studio 15 2017"**
  - GetFullPathNameW is defined in windows.h
  - Workaround to convince MSVC to allow redefinition of vsnprintf
  - Enforce utf-8 encoding by MSVC compiler switch (Fixes problem with Windows default code-page)
  - Add dll install path to PATH to allow tests to find the dlls (TODO: Allow tests after build, before install)
  - build jansson as a static library which is linked to dlite.dll
  - Fixing some dependencies

* **Python bindings**
  Difficult test on Windows, seems to be also an issue for other projects:
   https://github.com/pybind/pybind11/issues/2369
   https://github.com/numpy/numpy/issues/12667
  - Workaround: Let CMake generate a virtual environment for testing the Python-bindings (./bindings/python/CMakeLists.txts) - Windows, Linux
  - Add options WITH_PYHTON_BUILD_WHEEL and WITH_PYTHON_TEST_WHEEL for building and testing a python-wheel using an initially empty virtual environment - Windows, Linux
    Default to testing python-bindings only in the virtual environment on Windows
  - Remove AddDllDirectory etc. from dlite/\_\_init\_\_.py and instead copy the extensions
    directly into the package where _dlite.pyd will pick them up ("The current directory": https://docs.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order) - Windows
  - For Linux, do the same: Copy the *.so to the package directory. Add RPATH options to several targets with "$ORIGIN"/ so shared libraries find each other in the python-package ./dlite
  - Modify the setup.py such that it works on Windows and Linux (python setup.py install, python setup.py bdist_wheel) based on ideas from
https://github.com/galois-advertising/cmake_setup/blob/master/cmake_setup/cmake/__init__.py
  - Copy the plugins (*.dll or *.so) to the package directory ./dlite/share/dlite/storage-plugins. When importing dlite, modify os.environ['DLITE_ROOT'] such that it points to the package root directory. Not sure whether this will overlap with functionality intended for other environment variables such as PYTHON_STORAGE_PLUGINS

* **Not working**
  - Failed tests on Windows (Seems to be related to the yaml-plugin?)
    34 - test_mapping (Failed)  
    39 - test_yaml_storage (Failed)  
    40 - test_python_bindings-py_venv_python3.8.5 (Failed)  
    42 - test_entity-py_venv_python3.8.5 (Failed)  
  - Failed tests in Linux
    42 - test_python_bindings-py_venv_python3.8.5 (Failed)  
    44 - test_entity-py_venv_python3.8.5 (Failed)
  - hdf5 not compiled for Windows presently, but is hopefully similar to jansson

* **Documentation**
  - update the readme for the Windows-build, there is some duplicate documentation now in ./bootstrap-win.sh

* **Other**
  - The ./dlite/paths.py contains paths which are only relevant for the system where the code was compiled and may
    point to non-existing directories (i.e. When someone else is installing a pre-packaged wheel)
  - LD_LIBRARY_PATH and dlite_LD_LIBRARY_PATH have no special meaning on Windows
  - Building only the Release-Version for Windows


doc/build-with-vs.md
```diff
@@ -1,77 +1,67 @@
 Build with Visual Studio
```
jesper-friis, line 1: It would be great to be able to reproduce this instructions in a Windows docker worked on in PR #114.

jesper-friis (approved):
Great work. 

My only concern is that I am not completely sure if distributing several copies of the dlite DLLs in different directories is the optimal solution. @pscff, what is the reason for removing the calls to AddDllDirectory()? 

Anyway, everything seems to work and it is great to have a working setup.py for Windows, so I think we can merge to master and create a new issue addressing the possible duplication of dlls.

pscff:
Closing the pull-request since Windows-realted patches are now all in the windows-compile branch.

PR #120, opened 2021-07-07
Meta.sintef.no to onto ns.com

lovfall:


jesper-friis (approved):
Great work

Next step is to redirect http://onto-ns.com/meta to a entity server

PR #124, opened 2021-08-09
Builtin the JSON storage

jesper-friis:
A large PR that fixes many issues
- got rid of the jansson dependency - the JSON storage plugin is now always available (JSMN is now used for parsing json, while dlite itself implements serialising)
- json files with immutable entities are not modified by default
- fixed issue with loading collections
- fixed issue python interface to dlite search paths
- fixed several other bugs


jesper-friis:
Cross-compilation for Windows fails annoyingly. It seems that we are running into the following bug:
https://forum.winehq.org/viewtopic.php?f=8&t=33971

A workaround seems to downgrade wine

CMakeLists.txt
```diff
@@ -186,159 +180,20 @@ make_platform_paths(
...
-    endif()
-
+  find_package(HDF5 1.8.20 COMPONENTS C REQUIRED)
+  if(WIN32)
+    list(APPEND HDF5_LIBRARIES Kernel32)
+  else()
+    list(APPEND HDF5_LIBRARIES m dl)
   endif()
 else()
   set(HDF5_LIBRARIES "")
```
jesper-friis, line 194: See if we can get rid of this one

Dockerfile
```diff
@@ -137,7 +136,7 @@ RUN apt -qq update \
   && rm -rf /var/lib/apt/lists/*
 # Copy needed dlite files and libraries to slim image
 COPY --from=build /tmp/dlite-install /usr/local
-COPY --from=build /usr/lib/x86_64-linux-gnu/libjansson.so* /usr/local/lib/
+#COPY --from=build /usr/lib/x86_64-linux-gnu/libjansson.so* /usr/local/lib/
```
jesper-friis, line 139: Remove, not comment out

bindings/python/tests/test_entity.py
```diff
@@ -24,7 +24,7 @@
 assert not myentity.is_metameta
 
 # Store the entity to a new file
-myentity.save('json://xxx.json')
+myentity.save('json://xxx.json?mode=w')
```
jesper-friis, line 27: Also add checks with no mode

PR #125, opened 2021-08-10
Fixed syntax for CPack DEB Packages. 

jesper-friis:
- Added dependencies
- Removed user permissions from Docker production target

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -777,18 +777,15 @@ set(CPACK_SOURCE_IGNORE_FILES ${CPACK_SOURCE_IGNORE_FILES}
   ".git/*" "${dlite_BINARY_DIR}/*" "build*/*" "dist/*" "dlite.egg-info/*")
 
 # DEB package
-set(CPACK_DEBIAN_PACKAGE_DEPENDS
-  libhdf5
-  libjansson
-  python3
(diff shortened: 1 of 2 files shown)


jesper-friis:
Merged into #126

CMakeLists.txt
```diff
@@ -777,18 +777,15 @@ set(CPACK_SOURCE_IGNORE_FILES ${CPACK_SOURCE_IGNORE_FILES}
   ".git/*" "${dlite_BINARY_DIR}/*" "build*/*" "dist/*" "dlite.egg-info/*")
 
 # DEB package
-set(CPACK_DEBIAN_PACKAGE_DEPENDS
-  libhdf5
-  libjansson
-  python3
-  python3-numpy
-  )
+set(CPACK_DEBIAN_PACKAGE_DEPENDS "libhdf5-dev, libjansson-dev, python3-dev, python3-numpy, librdf0 libraptor2-dev")
```
jesper-friis, line 780: Is this correct cmake syntax? I would expect a space- (or semicolon)-separated list. 

The development packages are only needed for building. When installed as dependencies we only need the normal libraries.

An updated version of this branch is merged into PR #126 (to make the tests run through).

PR #126, opened 2021-08-15
Remove globals

jesper-friis:
All globals variables should now have been removed from dlite. dlite-mish implements now a globals handler that is shared with dynamically loaded plugins. This way we ensure that the plugins uses the same global state as the main program. 

Next step: when compiling the plugins, link dlite statically.
This should hopefully simplify compilation Windows.

### DIFF
--- a/src/dlite-codegen.c
+++ b/src/dlite-codegen.c
@@ -34,13 +34,6 @@ typedef struct {
   int use_native_typenames;
 } Globals;
 
-// /* Global variable indicating whether native typenames should be used.
-//    The default is to use portable type names. */
-// int dlite_codegen_use_native_typenames = 0;
-//
-// /* Pointer to template paths.  Access it via dlite_codegen_path_get(). */
-// static FUPaths *template_paths = NULL;
-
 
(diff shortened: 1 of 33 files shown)


src/dlite-codegen.c
```diff
@@ -13,6 +13,7 @@
 #include "dlite-macros.h"
 #include "dlite-codegen.h"
 
+#define GLOBALS_ID "dlite-codegen-globals-id"
```
quaat, line 16: The GLOBALS_ID are defined in multiple files with different values and get_globals have several reimplementations as well. Consider refactoring.

quaat (commented):
The duplicated code for declaring Globals and methods to work with them seems dangerous. In SOFT5 there is an example of how virtual functions can be implemented in C

src/dlite-codegen.c
jesper-friis: Yes, but all get_globals() are local to the file they are defined in and return different structures. Maybe it would be less confusing if we rename them according to the file their defined in. The same may be done with the GLOBALS_ID.

PR #137, opened 2021-09-10
Do not issue an error when testing if a json file exists

jesper-friis:


francescalb (approved):
This works as deisred now :)

I suggest to add
csv.meta.save('json:metadata.json')
in the try in  /examples/read_csv
to showcase metadata saving in json format

Issue #133, opened 2021-08-28
Same Blob metadata for all instances read with the blob storage plugin


jesper-friis:
When reading binary data using the blob storage plugin, dlite creates a corresponding metadata with a single property who's type is blob<N>, where <N> is the size of the data. This means that a new metadata is generated for each binary dataset with a new size.

By changing the type to uint8 and introducing a dimension corresponding to the size of the binary data, we just need a single Blob metadata.

Sub task:
* Distribute a json representation of the Blob metadata with dlite (in $DLITE_ROOT/share/dlite/entities/)
* Add $DLITE_ROOT/share/dlite/entities/*.json to the default storage search path

PR #141, opened 2021-09-14
Common blob metadata

jesper-friis:
Fixing issue #133

### DIFF
--- a/storages/python/tests/test_blob_storage.c
+++ b/storages/python/tests/test_blob_storage.c
@@ -1,0 +1,50 @@
+#include <stdlib.h>
+#include <stdio.h>
+
+#include "minunit/minunit.h"
+
+#include "utils/integers.h"
+#include "dlite.h"
+#include "dlite-macros.h"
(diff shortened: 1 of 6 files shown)


storages/python/tests/test_blob_storage.c
```diff
@@ -0,0 +1,50 @@
...
+
+
+DLiteInstance *inst = NULL;
+
+
+MU_TEST(test_load)
+{
+  char *url = "blob://" STRINGIFY(CURRENT_SOURCE_DIR) "/test_blob_storage.c";
+  inst = dlite_instance_load_url(url);
+}
```
TorgeirUstad, line 19: It would be natural to run mu_check(inst) here, since test_load tests loading of the blob. If the loading was unsuccessful, that failure should fail the test_load test itself, not just the later attempt to save the blob.

jesper-friis: Good point. Done

TorgeirUstad (approved):

PR #145, opened 2021-09-22
Issue128 semantic naming of functions

jesper-friis:


TorgeirUstad (commented):
When changing the arguments from "int n" to "int index" in the function signatures, you should also update the functions' comments, where it still says things like "path number 'n'".

This is a minor thing, but if one decides to change names, one should be consistent. :)

PR #152, opened 2021-09-22
Fix issue151

pscff:
Workaround using PYTHONHOME for embedding the Anaconda interpreter.


jesper-friis (approved):
Looks good. I am able to reproduce

PR #155, opened 2021-09-25
Issue154 fix test storage

jesper-friis:
Added workaround for issue #154 to ensure that the test_storage test does not segfaults.

Add a new issue for fix the actual problem with FreeLibrary

Note, this fix build on top of branch fix_issue151 (from PR #152).


pscff:
The fix works for me.

PR #158, opened 2021-09-25
Improved loading of json

jesper-friis:
`test_json_storage` test fails on MSVC.

This commit should fix the case where a json storage provides an
entity with the uri as its json identifier, as

    "dlite/1/test-c": {
      "meta": "http://onto-ns.com/meta/0.3/EntitySchema",
      "description": "some entity",
      "dimensions": [...],
      "properties": [...]
    },
    ...

The identifier is now correctly assigned to the uri of the entity.


pscff (approved):
Fixes the test_json_storage test on Windows. Tests on Linux pass. Crosscompile fails (workaround is in branch https://github.com/SINTEF/dlite/tree/workaround-for-mingw)

PR #161, opened 2021-09-27
Fix failing tests when cross-compiling with MinGW

jesper-friis:
Added two workarounds
* Copy all DLLs to the dlite Python package in binary dir, such
  that they can be found by the Python tests.
* Strip off initial "Z:" from paths when cross-compiling with MinGW.


pscff (approved):
Good. Works for the Github action.

PR #167, opened 2021-10-14
WIP OPTIMADE example

CasperWA:
Adding an example for the OPTIMADE API.

This initial example downloads a structure from the *odbx* database, specifically utilizing its OPTIMADE API.
Then a DLite instance is instantiated and populated from this datum.


CasperWA:
I think this is so outdated that this can be closed.
Furthermore, OPTIMADE support for DLite has been implemented in [OTEAPI-OPTIMADE](https://github.com/SINTEF/oteapi-optimade), including OTEAPI pipeline strategies.

PR #177, opened 2021-11-22
Python bindings to inspect dlites internal storage

pscff:
Adds a function to dlite-entity.c and the respective Python bindings to inspect the internal storage of dlite. Adds a Python-test.

Example from Python

```
import dlite
print(dlite.istore_get_uuids())
```

Returns a list of uuids available in the internal storage. Primarily intended for debugging/testing.

### DIFF
--- a/bindings/python/dlite-entity.i
+++ b/bindings/python/dlite-entity.i
@@ -79,6 +79,15 @@ bool dlite_swig_has_instance(const char *id, bool check_storages)
   return (dlite_instance_has(id, check_storages)) ? 1 : 0;
 }
 
+/* Returns list of uuids from istore. */
+char** dlite_swig_istore_get_uuids()
+{
+  int n;
+  char** uuids;
+  uuids = dlite_istore_get_uuids(&n);
+  return uuids;
+}
+
 %}
 
 
@@ -492,6 +501,13 @@ in the storage plugin path.
 %rename(has_instance) dlite_swig_has_instance;
 bool dlite_swig_has_instance(const char *id, bool check_storages=true);
 
+
+%feature("docstring", "\
+Returns a list of in-memory stored ids.
+
+") dlite_swig_istore_get_uuids;
+%rename(istore_get_uuids) dlite_swig_istore_get_uuids;
+char** dlite_swig_istore_get_uuids();
 
 %rename(_get_property) dlite_swig_get_property;
 %rename(_set_property) dlite_swig_set_property;
--- a/bindings/python/tests/global_dlite_state_mod2.py
+++ b/bindings/python/tests/global_dlite_state_mod2.py
@@ -1,0 +1,24 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+import os
+import pickle
+
+import dlite
+from dlite import Instance, Dimension, Property, Relation
+
+assert len(dlite.istore_get_uuids()) == 3 + 3
+
+thisdir = os.path.abspath(os.path.dirname(__file__))
+
+url = 'json://' + thisdir + '/MyEntity.json'
+
+# myentity is already defined via test_global_dlite_state, no new instance is added to istore
+myentity = Instance(url)
+assert myentity.uri == "http://onto-ns.com/meta/0.1/MyEntity"
+assert len(dlite.istore_get_uuids()) == 3 + 3
+
+i1 = Instance(myentity.uri, [2, 3], 'myid')
+assert i1.uri == "myid"
+assert i1.uuid in dlite.istore_get_uuids()
+assert len(dlite.istore_get_uuids()) == 3 + 4
+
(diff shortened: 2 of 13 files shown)


pscff:
Test of python-bindings fails for the Cross-compiled version.

pscff:
Removed testing of dlites global state from test_python_bindings and test the global state in a dedicated test. Using test_global_dlite_state in two different contexts (stand-alone plus from within test_python_bindings) was confusing.

bindings/python/tests/global_dlite_state_mod3.py
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+import os
+import pickle
```
jesper-friis, line 4: is pickle needed?

pscff: ..., probably left over from the original file and not needed

jesper-friis, line 4: No need to import pickle as far as I can see
```suggestion
```

bindings/python/tests/global_dlite_state_mod4.py
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+import os
+import pickle
```
jesper-friis, line 4: same here
```suggestion
```

bindings/python/tests/global_dlite_state_mod2.py
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+import os
+import pickle
```
jesper-friis, line 4:
```suggestion
```

bindings/python/dlite-entity.i
```diff
@@ -493,6 +502,13 @@ in the storage plugin path.
 bool dlite_swig_has_instance(const char *id, bool check_storages=true);
 
 
+%feature("docstring", "\
+Returns a list of in-memory stored ids.
+
+") dlite_swig_istore_get_uuids;
+%rename(istore_get_uuids) dlite_swig_istore_get_uuids;
+char** dlite_swig_istore_get_uuids();
```
jesper-friis, line 510: This will leak the memory returned by `dlite_swig_istore_get_uuids()` when the wrapped. 

We need to tell SWIG that is has ownership of the memory and that it should free it when the corresponding python object goes out of scope. Can be done with the SWIG `%newobject` and `%delobject` directives.

I think it is fine to fix this in a separate PR to not delay this one unnessesary

bindings/python/tests/test_python_bindings.py
```diff
@@ -32,9 +32,10 @@ def __repr__(self):
 
 
 def test(verbosity=1, stream=sys.stdout):
-    tests = [test for test in glob(os.path.join(thisdir, '*.py'))
-             if not test.endswith('__.py') and
-             not test.endswith('test_python_bindings.py')]
+    tests = [test for test in sorted(glob(os.path.join(thisdir, 'test_*.py')))
+             if not test.endswith('__.py')
+             and not test.endswith('test_python_bindings.py')
+             and not test.endswith('test_global_dlite_state.py')]
```
jesper-friis, line 38: Why do you exclude test_global_dlite_state.py?

src/dlite-entity.c
```diff
@@ -326,6 +326,34 @@ size_t dlite_instance_size(const DLiteMeta *meta, const size_t *dims)
...
+    instance_map_t* istore = _instance_store();
+    assert(istore);
+    const char* uuid;
+    map_iter_t iter;
+
+    iter = map_iter(istore);
+    *nuuids = 0;
+    while ((uuid = map_next(istore, &iter))) (*nuuids)++;
+
+    char** uuids;
```
jesper-friis, line 342:
```suggestion
    char** uuids=NULL;
```

jesper-friis (commented):
Nice feature. Thanks

bindings/python/tests/test_python_bindings.py
pscff: Because the test would be run in two different contexts and the global dlite state (istore) is then different making test assertions awkward. When run in test_python_bindings.py, previous tests have already populated the istore in ways which were not easy to understand (Because the other tests do not care about what they might populate the istore with, which is OK)

jesper-friis: Good point. Lets add a comment about that
```suggestion
             # Exclude test_global_dlite_state.py since the global state
             # that it is testing depends on the other tests.
             and not test.endswith('test_global_dlite_state.py')]
```

jesper-friis (approved):
Nice feature. Approved

PR #184, opened 2021-12-08
Bson plugin

TorgeirUstad:
Add BSON plugin for DLite

DLite seems to currently mess up binary strings, by replacing '\\' with '\\\\', which messes up control sequences (e.g. '\n' (hex '0a') becomes '\\\\n', which is interpreted as '\\\\' and 'n' (i.e hex '5c6e')). The plugin is therefore written as a workaround. If this issue can be resolved, the plugin Python code can probably be simplified.

The plugin has only been tested for BSON files generated by itself (by pymongo), not for BSON files generated by other sources.

EDIT: The problem mentioned above has been fixed, so the plugin has been simplified.

### DIFF
--- a/storages/python/python-storage-plugins/bson.py
+++ b/storages/python/python-storage-plugins/bson.py
@@ -1,0 +1,170 @@
+"""A DLite storage plugin for BSON written in Python."""
+import os
+import subprocess
+import sys
+
+import binascii
+from binascii import unhexlify
+import bson as pybson # Must be pymongo.bson
+
+import dlite
+from dlite.options import Options
+from dlite.utils import instance_from_dict
+
+
+class bson(DLiteStorageBase):
+    """DLite storage plugin for BSON."""
+
+    def open(self, uri, options=None):
+        """Open `uri`.
+
+        The `options` argument provies additional input to the driver.
+        Which options that are supported varies between the plugins.
+        It should be a valid URL query string of the form:
+
+            key1=value1;key2=value2...
+
+        An ampersand (&) may be used instead of the semicolon (;).
+
+        Typical options supported by most drivers include:
+        - mode : append (default) | r | w
+            Valid values are:
+            - append   Append to existing file or create new file
+            - r        Open existing file for read-only
+            - w        Truncate existing file or create new file
+
+        After the options are passed, this method may set attribute
+        `writable` to True if it is writable and to False otherwise.
+        If `writable` is not set, it is assumed to be True.
+        
+        The BSON data is translated to JSON.
+        """
+        self.options = Options(options, defaults='mode=append')
+        self.mode = dict(r='rb', w='wb', append='rb+')[self.options.mode]
+        self.writable = False if 'rb' in self.mode else True
+        self.uri = uri
+        self.d = {}
+        if self.mode in ('rb', 'rb+') and os.path.exists(uri):
+            with open(uri, self.mode) as f:
+                bson_data = f.read()
+                if pybson.is_valid(bson_data):
+                    self.d = pybson.decode(bson_data)
+                    if self.d == {}:
+                        raise EOFError("Failed to read BSON data from " + uri)
+                else:
+                    raise EOFError("Invalid BSON data in source " + uri)
+
+    def close(self):
+        """Close this storage and write the data to file.
+
+        Assumes the data to store is in JSON format.
+        """
+        if self.writable:
+            if self.mode == 'rb+' and not os.path.exists(self.uri):
+                mode = 'wb'
+            else:
+                mode = self.mode
+            esc_seq = { # Escape sequences
+                '5c30' : '00', # Null: '\\0' -> '\0' = hex 00
+                '5c61' : '07', # Bell: '\\a' -> '\a' = hex 07
+                '5c62' : '08', # Backspace: '\\b' -> '\b' = hex 08
+                '5c74' : '09', # Horizontal Tab: '\\t' -> '\t' = hex 09
+                '5c6e' : '0a', # Line Feed: '\\n' -> '\n' = hex 0a
+                '5c76' : '0b', # Vertical Tab: '\\v' -> '\v' = hex 0b
+                '5c66' : '0c', # Form Feed: '\\f' -> '\f' = hex 0c
+                '5c72' : '0d', # Carriage Return: '\\r' -> '\r' = hex 0d
+                '5c65' : '1b', # Escape: '\\e' -> '\e' = hex 1b
+                '5c5c' : '5c', # Backslash: '\\\\' -> '\\' = hex 5c
+                }
+            entries = {}
+            for uuid in self.queue():
+                d = str(self.d[uuid])
+                sp = d.split("bytearray(b'")
+                d = sp.pop(0)
+                # Convert binary strings in sp to hex strings
+                for s in sp:
+                    n = s.find("')")
+                    raws = "r'" + s[:n] + "'"
+                    # NOTE 02.12.2021:
+                    # DLite corrupts binary strings in Python, but not
+                    # raw strings, so s[:n].encode('utf-8').hex()
+                    # does not work, but 'hexs' below does
+                    hexs = raws.encode('utf-8').hex()
+                    # hexs must be '7227' + h + '27', where h is a
+                    # hexadecimal string of length divisible by two,
+                    # since each binary value maps to a two-digit
+                    # hexadecimal number.
+                    # Each pair of hexadecimal digits HH in h are
+                    # on one of the following forms:
+                    # > '5cXX' = an escape sequence in esc_seq,
+                    #            e.g. '5c6e' = b'\n'
+                    # > 'XXXX' = two binary characters,
+                    #            e.g. '3130' = b'10'
+                    # > '5c78' = the first four of eight digits
+                    #            representing one binary character
+                    #            on the form '5c78XXXX',
+                    #            e.g. '5c783130' = b'\x10'
+                    hexs = hexs.lstrip('7227') # Remove "r'"
+                    hexs = hexs.rstrip('27') # Remove "'"
+                    d = d + "'"
+                    pos = 0
+                    while pos < len(hexs):
+                        wrd = hexs[pos:(pos + 4)]
+                        pos = pos + 4
+                        if wrd[0:2] == '5c':
+                            if wrd[2:4] == '78':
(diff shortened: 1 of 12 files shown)


storages/python/python-storage-plugins/bson.py
```diff
@@ -0,0 +1,170 @@
...
+        If `writable` is not set, it is assumed to be True.
+        
+        The BSON data is translated to JSON.
+        """
+        self.options = Options(options, defaults='mode=append')
+        self.mode = dict(r='rb', w='wb', append='rb+')[self.options.mode]
+        self.writable = False if 'rb' in self.mode else True
+        self.uri = uri
+        self.d = {}
+        if self.mode in ('rb', 'rb+') and os.path.exists(uri):
```
jesper-friis, line 47: This will not fail if `uri` doesn't exists - but we want it to fail in this case since we are opening in read mode

storages/python/python-storage-plugins/bson.py
```diff
@@ -0,0 +1,170 @@
...
+            with open(uri, self.mode) as f:
+                bson_data = f.read()
+                if pybson.is_valid(bson_data):
```
jesper-friis, line 50: We can end the with-block here, since are done reading the data

storages/python/python-storage-plugins/bson.py
```diff
@@ -0,0 +1,170 @@
...
+                    self.d = pybson.decode(bson_data)
+                    if self.d == {}:
```
jesper-friis, line 52:
```suggestion
                    if not self.d:
```

storages/python/python-storage-plugins/bson.py
```diff
@@ -0,0 +1,170 @@
...
+                    d = d + "'" + s[(n + 2):]
+                entries[uuid] = d
+            with open(self.uri, mode) as f:
+                f.write(pybson.encode(entries))
+
+    def load(self, uuid):
+        """Load `uuid` from current storage and return it
+        as a new instance.
+        """
+        for key, value in self.d.items():
```
jesper-friis, line 132: No need to loop through `self.d`. You can get the value with `value = self.d[uuid]`

storages/python/python-storage-plugins/bson.py
```diff
@@ -0,0 +1,170 @@
...
+            if key == uuid:
+                if type(value) == str:
+                    value = eval(value) # Translate to dict
```
jesper-friis, line 135: This is a security issue. Are we 100% sure that the `value` cannot be a malicious designed string?
Eval can execute anything. Safer to import `ast` and write `value = ast.literal_eval(value)`.

storages/python/python-storage-plugins/bson.py
```diff
@@ -0,0 +1,170 @@
...
+                # Create instance and return it
```
jesper-friis, line 136: Isn't it possible to just return `instance_from_dict(value)` at this point?

storages/python/python-storage-plugins/bson.py
```diff
@@ -0,0 +1,170 @@
...
+                        raise EOFError("Failed to read BSON data from " + uri)
+                else:
+                    raise EOFError("Invalid BSON data in source " + uri)
+
+    def close(self):
+        """Close this storage and write the data to file.
+
+        Assumes the data to store is in JSON format.
+        """
+        if self.writable:
```
jesper-friis, line 62: I am probably an ignorant, but I don't understand why we need these conversions.

I testet the following which works for me (prints True):
```python
import bson

import dlite
from dlite.utils import instance_from_dict


inst = dlite.get_instance(dlite.ENTITY_SCHEMA)

data = bson.encode(inst.asdict())
with open('raw.bson', 'wb') as f:
    f.write(data)

with open('raw.bson', 'rb') as f:
    newdata = f.read()
d = bson.decode(newdata)
inst2 = instance_from_dict(d)

print(inst == inst2)
```

storages/python/tests/test_bson_storage.c
```diff
@@ -0,0 +1,110 @@
...
+{
+  DLiteStorage* s = NULL;
+
+  // Load JSON metadata
+  char* url = "json://" STRINGIFY(DLITE_ROOT) "/src/tests/test-entity.json?mode=r";
+  DLiteInstance* json_meta = (DLiteInstance*)dlite_meta_load_url(url);
+  char* json_str1 = dlite_json_aprint(json_meta, 0, 1);
+  while (dlite_instance_decref(json_meta)); // Remove all references to json_meta
+
+  // Load BSON metadata
```
jesper-friis, line 52:
```suggestion
  // Load BSON data
```

storages/python/tests/test_bson_storage.c
```diff
@@ -0,0 +1,110 @@
...
+  bson_str1 = dlite_json_aprint(bson_data, 0, 1);
+  while (dlite_instance_decref(bson_data)); // Remove all references to bson_data
+  bson_data = dlite_instance_load(s, "e076a856-e36e-5335-967e-2f2fd153c17d");
+  mu_check(bson_data);
+  char* bson_str2 = dlite_json_aprint(bson_data, 0, 1);
+  while (dlite_instance_decref(bson_data)); // Remove all references to bson_data
+
+  // Compare JSON and BSON data
+  mu_assert_int_eq(0, strcmp(json_str1, bson_str1));
+  mu_assert_int_eq(0, strncmp(json_str2, bson_str2, 47));
```
jesper-friis, line 86: why do you use strncmp and not just strcmp?

storages/python/tests/test_bson_storage.c
```diff
@@ -0,0 +1,110 @@
...
+#include "dlite.h"
+#include "dlite-macros.h"
+#include "dlite-storage-plugins.h"
+
+MU_TEST(test_save)
+{
+  DLiteStorage* s = NULL;
+
+  // Load JSON metadata
+  char* url = "json://" STRINGIFY(DLITE_ROOT) "/src/tests/test-entity.json?mode=r";
```
jesper-friis, line 12: In general we would like to keep lines shorter than 80 characters - makes them easy to read in a 80 character wide termical...

storages/python/python-storage-plugins/bson.py
```diff
@@ -0,0 +1,170 @@
...
+        self.writable = False if 'rb' in self.mode else True
+        self.uri = uri
+        self.d = {}
+        if self.mode in ('rb', 'rb+') and os.path.exists(uri):
+            with open(uri, self.mode) as f:
+                bson_data = f.read()
+                if pybson.is_valid(bson_data):
+                    self.d = pybson.decode(bson_data)
+                    if self.d == {}:
+                        raise EOFError("Failed to read BSON data from " + uri)
```
jesper-friis, line 53: It is good practice to keep lines below 80 characters

storages/python/tests/test_bson_storage.c
jesper-friis: Ahh, ser nå hvorfor. bson-parseren setter ikke `uri`. Det er vel egentlig noe vi ønsker at den skal gjøre.

jesper-friis (commented):
Hi, I have added some comments alongside the code

storages/python/tests/test_bson_storage.c
TorgeirUstad: I know, but I saw that it wasn't practiced throughout the DLite source code, so I didn't pay attention to it. But I'll fix it.

storages/python/python-storage-plugins/bson.py
TorgeirUstad: I know. I looked at ast.literal_eval(), but decided against using it due to the warning:
https://docs.python.org/3/library/ast.html?highlight=ast%20literal_eval#ast.literal_eval

I'm not sure if the issue stated in the warning would be an issue or not in practical use of the plugin.

storages/python/python-storage-plugins/bson.py
TorgeirUstad: I'll fix it (see reply above).

storages/python/python-storage-plugins/bson.py
TorgeirUstad: Is this metadata? The problem I encountered was with actual data (the 'myblob' properties in the file 'src/tests/test-data.json'). The DLite entity stores e.g. the 'myblob' value '0a0010' (for uuid 'e076a856-e36e-5335-967e-2f2fd153c17d') as "bytearray(b'\n\x00\x10')", and bson.encode() didn't accept that input, so it had to be translated. But translating it directly back to '0a0010' didn't work**, as DLite somehow caused the backslashes to be doubled, so that the translated hexadecimal value became '5c6e5c7830305c783130' (corresponding to b'\\\n\\\x00\\\x10').

** It worked when run through the Python interpreter, but not when run through DLite. This is why I wrote "The plugin is therefore written as a workaround. If this issue can be resolved, the plugin Python code can probably be simplified."

storages/python/python-storage-plugins/bson.py
TorgeirUstad: For metadata, yes, For actual data, it turned out that I had to handle the hexadecimal input explicitly.

storages/python/python-storage-plugins/bson.py
TorgeirUstad: Ok. Fail, as in raising an exception?

storages/python/python-storage-plugins/bson.py
jesper-friis: You are right, `ast.literal_eval()` does not protect against denial-of-service attacks, but it is much better than `eval()`. I definitely think you should use ast.literal_eval().

storages/python/python-storage-plugins/bson.py
jesper-friis: Yes, raising an exception with a good message would be the preferred behaviour

storages/python/python-storage-plugins/bson.py
jesper-friis: Ahh, I see. I think the right solution would be to fix `instance_from_dict()` rather than you having to handle all the edge cases. Please add an issue.

storages/python/tests/test_bson_storage.c
jesper-friis: Probably because there have been several authors. I think we should standardise on 80 character line length.

jesper-friis (approved):

PR #187, opened 2021-12-18
Read blob data from Python

jesper-friis:
* Added support for reading any Python object that supports the buffer protocol as blob data.
* Also support reading Python hex-encoded strings as blob data.

### DIFF
--- a/bindings/python/dlite-python.i
+++ b/bindings/python/dlite-python.i
@@ -214,7 +214,7 @@ int dlite_swig_read_python_blob(PyObject *src, uint8_t *dest, size_t n)
       FAIL("failed preparing string");
     len = (long)PyUnicode_GET_LENGTH(src);
     if (len != (long)n*2)
-      FAIL3("cannot convert Python string of lenght %d (expected length %d) "
+      FAIL3("cannot convert Python string of length %d (expected length %d) "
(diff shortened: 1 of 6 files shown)


TorgeirUstad (changes requested):
### strhex_encode():
The returned value doesn't represent the amount of data actually written to 'hex'. In every iteration of the for loop, the method either returns on line 209 of 'strutils.c' with a negative value, or n becomes the number of characters that one wants to write (i.e. two), regardless of whether the characters are written or not. So the returned value is either negative or 2 * size (cf. the tests in 'test_strutils.c').

### strhex_decode():
On line 243 of 'strutils.c', the second criterion must be "c <= 'f'", since 'g' through 'z' are not valid hexadecimal digits.

### dlite-python.i:
Typo on line 217 ("lenght").

jesper-friis:
Thanks for the review. The typo and bug in strhex_decode() are corrected. 

Regarding strhex_encode(), returning the number of characters one wants to write was actually the intended behaviour, mimicking the return value of the standard library snprintf().  I have updated the documentation to make it more clear. Or do you have a suggestion for a more sensible behaviour?

TorgeirUstad (approved):

Issue #146, opened 2021-09-22
Building without Python support fails


pscff:
One of the tests expects a reference which seems to be generated only with -DWITH_PYTHON=ON

```
$ mkdir out/build/noPython
$ cd out/build/noPython
$ cmake -G "Unix Makefiles" -DCMAKE_BUILD_TYPE:STRING="Release" -DWITH_DOC=OFF -DWITH_HDF5=OFF -DWITH_PYTHON=OFF -DALLOW_WARNINGS=ON ../../.. 
$ cmake --build . --config Release
...
[ 74%] Linking C executable test_mapping
/usr/bin/ld: ../libdlite.so: undefined reference to `dlite_python_mapping_next'
/usr/bin/ld: ../libdlite.so: undefined reference to `dlite_python_mapping_paths_get'
collect2: error: ld returned 1 exit status
make[2]: *** [src/tests/CMakeFiles/test_mapping.dir/build.make:86: src/tests/test_mapping] Error 1
make[1]: *** [CMakeFiles/Makefile2:1994: src/tests/CMakeFiles/test_mapping.dir/all] Error 2
make: *** [Makefile:163: all] Error 2
```

Obvious workaround

- Compile with Python support
- One (or more?) tests must be excluded when setting -DWITH_PYTHON=OFF

jesper-friis:
Hi @pscff, I guess this issue is fixed and can be closed?

pscff:
The issue is still present for the current master, the build fails already in the cmake-configure step now. How to reproduce: Set `-DWITH_PYTHON=OFF` in your build-system and try to build.

PR #197, opened 2022-01-08
Fixed issue #146 Building without Python support fails

jesper-friis:


pscff (approved):
Building and testing without Python works perfectly.

PR #198, opened 2022-01-08
Test dehydrogenation

jesper-friis:
Run the dehydrogenation example as part of the tests

This PR depends on PR #197. Do not merge before #197 has been merged.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -5,6 +5,7 @@ project(dlite-examples)
 if(WITH_PYTHON)
 set(python_examples
   read-csv
+  dehydrogenation
   )
 endif()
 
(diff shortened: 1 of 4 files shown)


examples/CMakeLists.txt
```diff
@@ -76,7 +79,8 @@ if(FORCE_EXAMPLES OR EXISTS ${CMAKE_INSTALL_PREFIX}/share/dlite/examples)
     test_example(ex4)
   endif()
   #test_example(ex5d)
-
+  if(WITH_PYTHON)
+  endif()
```
francescalb, line 83: Shouldn't this be on line 105, after the foreach-loop?

jesper-friis: Line 82-83 should simply be removed... well spotted

PR #199, opened 2022-01-08
The crosscompile CI test takes too long. Changed it to only run nightly

jesper-friis:


CasperWA (approved):
I don't know why the CI doesn't run currently on this PR as it should. However, this seems fine to me. I'd suggest maybe running this on pushes to `master` or on `pull_request` events still? But I'll leave that up to you. The ultimate goal is to make sure any new changes don't unknowingly destroy parts of the software. For this the `pull_request` event is nice. It will naturally still then trigger this workflow for every new push to a branch for which a PR has been opened, but then at least it will not run for branches where this is not the case. I'll leave it up to you though and will approve it now, since there is nothing inherently wrong with this change as I see it.

jesper-friis:
Good suggestion to also run it on `pull_request` and `push to master`

Is the yaml-syntax correct?

Issue #176, opened 2021-11-06
Add keyword argument support to SWIG-generated code


jesper-friis:
According to the documentation of SWIG 4.0, support for keyword argument can be added by providing the -keyword argument when running swig.

This cannot be combined with overloaded functions, so we have to fix all overloaded functions and update the Python interface to dlite.

jesper-friis:
Should probably be fixed together with issue #162 to minimise changes in the API.

PR #202, opened 2022-01-11
Swig keywords

kriwiik:
Initial fix to #176. Swig still gives warnings, so a new issue should be created to fix this.

Keyword support has been added for swig generated code. Since this cannot be combined with overloaded functions, we had to fix all overloaded functions and update the Python interface to dlite. In practice this meant that the Instance() constructor and all its calls throughout dlite had to be changed. Helper functions were created as class methods (see dlite-entity-python.i).


bindings/python/dlite-entity.i
```diff
@@ -257,16 +257,18 @@ Instance(driver, location, options, id=None)
     and `options`. `id` is the id of the instance in the storage (not
     required if the storage only contains more one instance).
 
-Instance(uri, dimensions, properties, iri, description)
+Instance(uri, dimensions, properties, description)
```
jesper-friis, line 260: We should remember to update the documentation.

In this case we should say that Instance() is now intended to be called from one of the new class methods

    create_from_metaid(cls, metaid, dims, id=None)
    create_from_url(cls, url, metaid=None)
    create_from_storage(cls, storage, id=None, metaid=None)
    create_from_location(cls, driver, location, options=None, id=None)
    create_metadata(cls, uri, dimensions, properties, description)

and refer to their documentation for details.

bindings/python/factory.py
```diff
@@ -1,16 +1,12 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 """A module that makes it easy to add dlite functionality to existing classes.
-
 Class customisations
```
jesper-friis, line 4:
```suggestion

Class customisations
--------------------
```

bindings/python/factory.py
```diff
@@ -218,4 +218,4 @@ def classfactory(theclass, meta=None, url=None, storage=None, id=None):
         __init__=BaseExtension.__init__
     )
 
-    return type(meta.name, (theclass, BaseExtension), attr)
+    return type(meta.name, (theclass, BaseExtension), attr)
```
jesper-friis, line 221: End files with newline
```suggestion
    return type(meta.name, (theclass, BaseExtension), attr)
    
```

jesper-friis (commented):
Great work! I just have three minor documentation and style-related suggestions.

jesper-friis (approved):
Very nice. Sorry for the late review.

Issue #182, opened 2021-12-07
Make DLiteStorageBase and DLiteMappingBase available from the python api


jesper-friis:
Make dlite storage and mapping plugin less magic by subclassing from `dlite.DLiteStorageBase` and `dlite.DLiteStorageBase` instead of magically add these classes to the namespace when loading a Python plugin.

Issue #138, opened 2021-09-12
Add python tests suite for reading/writing different storage types and file formats


jesper-friis:
Should also serve as a documentation.

francescalb:
add to: bindings/python/tests/test_storage.py

e.g.: csv.meta.save('json:metadata.json')

TorgeirUstad:
This test suite has been partially implemented in the branch 'Python test system for storages'. Tests for blobs and CSV files remain to be implemented.

PR #203, opened 2022-01-11
Python test system for storages

TorgeirUstad:
Add Python tests for DLite storage plugins (partial resolution of issue #138).

Tests have been added for BSON, PostgreSQL and YAML. Tests for blobs and CSV files remain to be added.

The PostgreSQL test avoids the need for a PostgreSQL database server. The plugin has been tested against such a server (and passed after a minor modification). The test works by testing against files rather than against a database. The loading tests use two database dump files as input, and the saving tests compare the SQL commands that would have been executed by a server, with those used to generate the instances contained in the two dump files.

CMake files have been added/modified to make these tests run with the existing test system. Thanks to Jesper for help with these files.


storages/python/python-tests/bson_test.py
```diff
@@ -0,0 +1,73 @@
...
+from run_python_tests import print_test_exception
+
+
+print('Running Python test <bson_test>...')
+thisdir = Path(__file__).absolute().parent
+input_path = thisdir / 'input'
+plugin = thisdir.parent / 'python-storage-plugins/bson.py'
+
+try:
+    with open(plugin, 'r') as orig:
```
jesper-friis, line 15: Instead of opening and changing the plugins, it is much easier to simply import them:

```python
plugins_dir = thisdir.parent / 'python-storage-plugins'
sys.path.insert(plugins_dir)

class DLiteStorageBase:
    """Dummy replacement for the built-in class in the Python plugins."""
    
import bson
```

storages/python/python-tests/bson_test.py
```diff
@@ -0,0 +1,73 @@
...
+    else:
+        raise ValueError('...Saving data failed!')
+    
+    print('Test <bson_test> ran successfully')
+except Exception as err:
+    if __name__ == '<run_path>':
+        print_test_exception(err)
+    else:
+        raise
+finally:
```
jesper-friis, line 68: No urgent need to clean up, since the preferred way to run the tests is via cmake/ctest. But it doesn't hurt either...

ctest will always run the tests from the build directory, which will ensure that any output from build or tests will go to build directory and avoid any potential interference between parallel builds...

storages/python/python-tests/postgresql_test.py
```diff
@@ -0,0 +1,275 @@
...
+            n = save_start + 2
+            lines[n - 1] = ind8 + 'ret = {"uuid": inst.uuid}\n'
+            while not lines[n].startswith(df + 'table_exists'):
+                lines[n] = lines[n].replace('self.conn', '#')
+                lines[n] = lines[n].replace('self.cur.execute(', \
+                    'ret = extract_exec_args(ret, ')
+                if lines[n].startswith(ind8 + 'if not self.table'):
+                    lines[n] = '\n'
+                    lines[n + 1] = '\n'
+                n = n + 1
```
jesper-friis, line 73:
```suggestion
                n += 1
```
Generally preferable, since it avoids generating a new object

storages/python/python-tests/postgresql_test.py
```diff
@@ -0,0 +1,275 @@
...
+            lines[n - 1] = ind8 + 'return ret\n'
+            
+            del lines[(load_start + 5):(save_start - 1)]
+            del lines[(close_start + 2):(load_start - 1)]
+            del lines[(open_start + 3):(close_start - 1)]
+            s = 'from postgresql_test import open_pgsql, load_pgsql, ' \
+                + 'extract_exec_args\n' + str().join(lines)
+            s = s.replace('postgresql(DLiteStorageBase)', \
+                          'dlite_postgresql(object)')
+        exec(s)
```
jesper-friis, line 83: Could be replaced with import

See similar comment for bson_test.py

storages/python/python-tests/bson_test.py
```diff
@@ -0,0 +1,73 @@
+"""Script to test the DLite plugin 'bson.py' in Python."""
```
jesper-friis, line 1: Test files conventionally starts with "test_". Tools like pytest or nose will automatically find all python tests in files matching "test_*.py". It also makes sorting more logically

storages/python/python-tests/run_python_tests.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+if __name__ == '__main__':
+    # The test scripts to run
+    tests = [
+        'bson_test.py',
+        'postgresql_test.py',
+        'yaml_test.py',
+        ]
+    path = str(os.getcwd()).replace('\\', '/') + '/'
```
jesper-friis, line 24:
```python
from pathlib import Path

path = Path('.')  # relative path
path = Path(__file__).absolute().parent  # absolute path
```
The above is preferred

storages/python/python-tests/yaml_test.py
```diff
@@ -0,0 +1,73 @@
...
+print('Running Python test <yaml_test>...')
+thisdir = Path(__file__).absolute().parent
+input_path = thisdir / 'input'
+plugin = thisdir.parent / 'python-storage-plugins/yaml.py'
+
+try:
+    with open(plugin, 'r') as orig:
+        s = orig.read()
+    s = s.replace('yaml(DLiteStorageBase)', 'dlite_yaml(object)')
+    exec(s)
```
jesper-friis, line 18: Same as commented on in the bson_test.py - simpler to use import

jesper-friis (commented):
Great work. I added some comments

CMakeLists.txt
```diff
@@ -133,7 +133,7 @@ set(DLITE_TEMPLATE_DIRS              "share/dlite/templates")
 set(DLITE_STORAGE_PLUGIN_DIRS        "share/dlite/storage-plugins")
 set(DLITE_MAPPING_PLUGIN_DIRS        "share/dlite/mapping-plugins")
 set(DLITE_PYTHON_MAPPING_PLUGIN_DIRS "share/dlite/python-mapping-plugins")
-set(DLITE_PYTHON_STORAGE_PLUGIN_DIRS "share/dlite/python-storage-plugins")
+set(DLITE_PYTHON_STORAGE_PLUGIN_DIRS "share/dlite/python_storage_plugins")
```
jesper-friis, line 136: Changing one directory path to underscore instead of hyphen looks strange. 

For Python 3.5+ you can use the following to import a given file as a python module. This also avoids messing with sys.path
```python
import importlib.util
spec = importlib.util.spec_from_file_location("module.name", "/path/to/file.py")
mod = importlib.util.module_from_spec(spec)
spec.loader.exec_module(mod)
mod.MyClass()
```

bindings/python/dlite-python.i
```diff
@@ -1296,4 +1296,12 @@ PyObject *_get_DLiteError(void);
 
 %pythoncode %{
   DLiteError = _dlite._get_DLiteError()
+  
+  class DLiteStorageBase:
```
jesper-friis, line 1300: These should be done properly, as done for DLiteError above - fixes issue #182

jesper-friis: I pushed an implementation for proper access to the underlying C objects for DLiteStorageBase and DLiteMappingBase to PR #208

PR #210, opened 2022-02-03
Make workflow3 run in dehydrogenation example.

francescalb:
Changed mapsTo from rdf:Property to owl:AnnotationProperty.

Protege and Owlready2 (and therefore also EMMOntoPy) are based on owl as
base ontology and they change the ontology when loading the original
mapsTo. This is not very practical. We therefore changed the ontology
instead of updating EMMOntoPy. This might be changed in the future.
Commented in the README.md.

Note that we have also made mapsTo an rdfs:subPropertyOf rdfs:SubClassOf. This adds semantic meaning, but is not used for anything right now.


jesper-friis (approved):
:+1:

PR #213, opened 2022-02-06
Fixed deprecated old cmake CMP0078 policy

jesper-friis:
The old behaviour of the CMake CMP0078 policy has long been marked as deprecated. After upgrading cmake to 3.22, I get warnings about the old behaviour will be removed in the next release. This PR fixes this issue + two other small issues related to  upgrading cmake and Python. Backward compatibility should be retained.
- Use the new behaviour of cmake CMP0078 policy
- Replaced use of deprecated sysconfig Python package with the site package
- Removed deprecated COLS_IN_ALPHA_INDEX doxygen variable
- Cleaned up bindings/python/CMakeLists.txt

### DIFF
--- a/bindings/python/CMakeLists.txt
+++ b/bindings/python/CMakeLists.txt
@@ -16,6 +16,7 @@ else()
   configure_file(__init__inbuild.py.in __init__.py)
 endif()
 
+
 # Determine size of size_t
 include(CheckTypeSize)
 check_type_size(size_t SIZEOF_SIZE_T)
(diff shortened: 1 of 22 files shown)


bindings/python/CMakeLists.txt
```diff
@@ -74,18 +63,21 @@ swig_add_library(dlite
...
-swig_link_libraries(dlite
+swig_link_libraries(dlitepy
   dlite-utils
   dlite
   ${Python3_LIBRARIES}
   )
 
 
+
+
```
CasperWA, line 80:
```suggestion
```

CasperWA (changes requested):
Cheers @jesper-friis

Just some extra empty lines that should be removed. Then all good to go.

CasperWA (approved):

PR #216, opened 2022-02-12
Fixed documentation links - doxygen doesn't like hyphens

jesper-friis:


TorgeirUstad:
I wasn't able to generate the Doxygen documentation from `Doxyfile.in`. It seems that the variables `@dlite_SOURCE_DIR@`, `@dlite_BINARY_DIR@`, `@dlite-src_BINARY_DIR@` and `@dlite-src-utils_BINARY_DIR@` are causing problems. I used Doxygen manually, so maybe it would have worked better if I had used a makefile (I see that the main `README` says to run `make doc`, but I guess that requires that one has a proper makefile, which I didn't have).

lovfall:
> I wasn't able to generate the Doxygen documentation from `Doxyfile.in`. It seems that the variables `@dlite_SOURCE_DIR@`, `@dlite_BINARY_DIR@`, `@dlite-src_BINARY_DIR@` and `@dlite-src-utils_BINARY_DIR@` are causing problems. I used Doxygen manually, so maybe it would have worked better if I had used a makefile (I see that the main `README` says to run `make doc`, but I guess that requires that one has a proper makefile, which I didn't have).

Documentation builds out of the box on my linux distro. Cmake generate doc makefile as expected

lovfall (approved):
This looks good. I can confirm that the links work. I assume the actual documentation will be filled in at a later stage for the templates added.

PR #224, opened 2022-03-06
Fixed bug in the python API for storage.save()

jesper-friis:


francescalb (changes requested):
Good! Works for me locally, but needs a test for approval.

Suggestion something like
inst = dlite.Instance.create_from_url(f'json://{path/to/entities}/entity.json')
storage = dlite.Storage(driver_or_url="json", location=f"{path_to_collections}/coll.json", options="")

inst.save_to_storage(storage)

PR #244, opened 2022-03-20
Fix 241

### LINKED ISSUE #241: Python-plugin to accept Posix paths and not only strings
have to convert path to string with Instance.create_from_url(path)

jesper-friis:
fix #241

Made it possible use pathlib.Path objects as argument several places.


bindings/python/dlite-entity-python.i
```diff
@@ -375,10 +375,28 @@ def get_instance(id: "str", metaid: "str"=None, check_storages: "bool"=True) ->
...
+        """
+        if isinstance(dest, Storage):
+            self.save_to_storage(storage=dest)
+        elif location:
+            with Storage(dest, str(location), options) as storage:
+                storage.save(self)
+        elif isinstance(dest, str):
+            self.save_to_url(dest)
+        else:
+            raise DLiteError('arguments does not match any call signature')
```
francescalb, line 398:
```suggestion
            raise DLiteError('Arguments do not match any call signature')
```

francescalb (approved):
Approved, but fix typo in comment first.

PR #248, opened 2022-04-03
Added option `force` to replace an existing instance in a collection.

### LINKED ISSUE #223: Ensure that collection labels are unique
Make it an error if one adds an instance with an already existing label.

In the Python interface, consider to add an option `force=False` to replace a possible existing label. Remember to decref the replaced instance.

jesper-friis:
close #223 

It is already an error in the C api (and therefore also from Python without `force`) to add an instance with an existing label.


CasperWA (approved):
Looks good to me

PR #258, opened 2022-05-01
Changed default JSON serialisation from Python to SOFT7

jesper-friis:


quaat (commented):
@CasperWA has prepared a Pydantic datamodel for working with SOFT7, which is superior to OrderedDicts: https://github.com/SINTEF/soft7/blob/main/s7/pydantic_models/soft7.py
It could be an idea to incorporate this.

jesper-friis:
> @CasperWA has prepared a Pydantic datamodel for working with SOFT7, which is superior to OrderedDicts: https://github.com/SINTEF/soft7/blob/main/s7/pydantic_models/soft7.py It could be an idea to incorporate this.

Good idea. But this PR is about formatting the JSON/YAML representation according to SOFT7. Would it make sense to create a new PR that add the methods `dlite.Instance.from_soft7()` and `dlite.Instance.as_soft7()`?  The former should only require pydantic to be installed, while the latter would also require SOFT7.

PR #259, opened 2022-05-02
WIP: fix read yaml

### LINKED ISSUE #165: yaml-plugin: read_yaml
if you try to read a yaml-file with data (e.g. faithful.yaml) into an instance 

```
inst_yaml = dlite.Instance(f'yaml:{yamlfile}?'
                     'mode=r')
```

it does not work properly as it expects an uuid; however the uuid should be generated during the ingest

see yaml_plugin.py

```
    def load(self, uuid):
        """Loads `uuid` from current storage and return it as a new instance."""
        uuid = dlite.get_uuid(uuid)
        return instance_from_dict(self.d[uuid])
```
something like this fixes the problem

```
    def load(self, uuid=None):
        """Loads `uuid` from current storage and return it as a new instance."""
        return instance_from_dict(self.d)

```

jesper-friis:
Closes #165

### DIFF
--- a/bindings/python/dlite-entity-python.i
+++ b/bindings/python/dlite-entity-python.i
@@ -203,7 +203,11 @@ def get_instance(id: "str", metaid: "str"=None, check_storages: "bool"=True) -> 
 
     # Override default generated __init__() method
     def __init__(self, *args, **kwargs):
+        errclr()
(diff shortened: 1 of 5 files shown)


jesper-friis:
Closing because issue #165 is fixed in PR #266

PR #54, opened 2021-01-03
WIP: Add transactions

jesper-friis:
Work in progress. Started adding transactions.

PR #260, opened 2022-05-02
Adding transactions

jesper-friis:
Implements transactions and replaces old PR #54 

A separate issue (#261) is created for adding support for proper serialisation/deserialisation of transactions.


bindings/python/tests/test_transaction.py
```diff
@@ -0,0 +1,73 @@
...
+else:
+    HAVE_PYTEST = True
+
+import dlite
+
+
+# Configure paths
+thisdir = Path(__file__).parent.absolute()
+
+#dlite.storage_path.append(thisdir / 'entities' / '*.json')
```
CasperWA, line 16:
```suggestion
```

bindings/python/tests/test_transaction.py
```diff
@@ -0,0 +1,73 @@
...
+dlite.storage_path.append(thisdir / '*.json')
+Person = dlite.get_instance('http://onto-ns.com/meta/0.1/Person')
+
+
```
CasperWA, line 20:
```suggestion
```

bindings/python/tests/test_transaction.py
```diff
@@ -0,0 +1,73 @@
...
+person.name = "Knud H. Thomsen"
+person.age = 30
+person.skills = ["writing", "humor", "history", "human knowledge"]
+
+# Update person 6 times and create a snapshot before each change
+for i in range(6):
+    person.snapshot()
+    person.age += 5
+
+assert person.age == 60
```
CasperWA, line 31:
```suggestion
assert person.age == 30 + 6 * 5  # 60
```

bindings/python/tests/test_transaction.py
```diff
@@ -0,0 +1,73 @@
...
+assert person.get_snapshot(0).age == 60
+assert person.get_snapshot(1).age == 55
+assert person.get_snapshot(6).age == 30
```
CasperWA, line 34:
```suggestion
for i in range(7):
    assert person.get_snapshot(i).age == 30 + (6 - i) * 5
```

bindings/python/tests/test_transaction.py
```diff
@@ -0,0 +1,73 @@
...
+if HAVE_PYTEST:
+    with pytest.raises(dlite.DLiteError):
+        person.get_snapshot(7)
+
```
CasperWA, line 38:
```suggestion
        person.get_snapshot(7)
else:
    try:
        person.get_snapshot(7)
    except dlite.DLiteError:
        pass
    else:
        raise Exception("Should've failed test (getting non-existant snapshot), but didn't".)
```

bindings/python/tests/test_transaction.py
```diff
@@ -0,0 +1,73 @@
...
+assert person._get_parent_uuid() == person.get_snapshot(1).uuid
+assert person._get_parent_hash() == person.get_snapshot(1).get_hash()
+
+# Chech mutability
```
CasperWA, line 42:
```suggestion
# Check mutability
```

bindings/python/tests/test_transaction.py
```diff
@@ -0,0 +1,73 @@
...
+assert person.is_frozen() == False
+assert person.get_snapshot(1).is_frozen() == True
+assert person.get_snapshot(2).is_frozen() == True
```
CasperWA, line 45: Other than checking `is_frozen()` value, you might also want to check that it truly _is_ frozen? I.e., checking that the `is_frozen()` value is respected.

bindings/python/tests/test_transaction.py
```diff
@@ -0,0 +1,73 @@
...
+
+assert inst.get_snapshot(0).age == 55
+assert inst.get_snapshot(1).age == 50
+assert inst.get_snapshot(2).age == 45
+assert inst.get_snapshot(3).age == 40
+assert inst.get_snapshot(4).age == 35
+assert inst.get_snapshot(5).age == 30
+if HAVE_PYTEST:
+    with pytest.raises(dlite.DLiteError):
+        inst.get_snapshot(6)
```
CasperWA, line 62:
```suggestion
        inst.get_snapshot(6)
else:
    try:
        inst.get_snapshot(6)
    except dlite.DLiteError:
        pass
    else:
        raise Exception("Should've failed test (getting non-existant snapshot), but didn't".)
```

bindings/python/tests/test_transaction.py
```diff
@@ -0,0 +1,73 @@
...
+inst.age = 50
+inst.snapshot()
+inst.age = 55
+
+assert inst.get_snapshot(0).age == 55
+assert inst.get_snapshot(1).age == 50
+assert inst.get_snapshot(2).age == 45
+assert inst.get_snapshot(3).age == 40
+assert inst.get_snapshot(4).age == 35
+assert inst.get_snapshot(5).age == 30
```
CasperWA, line 59:
```suggestion
assert inst.get_snapshot(0).age == 55
assert inst.get_snapshot(1).age == 50
for i in range(4):
    assert inst.get_snapshot(i + 2).age == person.get_snapshot(3 + i).age
```

bindings/python/tests/test_transaction.py
```diff
@@ -0,0 +1,73 @@
...
+if HAVE_PYTEST:
+    with pytest.raises(dlite.DLiteError):
+        inst.get_snapshot(6)
+
+#print(inst)
+
```
CasperWA, line 65:
```suggestion
```

bindings/python/tests/test_transaction.py
```diff
@@ -0,0 +1,73 @@
...
+# Push to storage
+db = thisdir / "output" / "db.json"
+if db.exists():
+    db.unlink()  # Make sure that the db is empty
+
+with dlite.Storage("json", db, "mode=w") as storage:
+    #person.push_snapshot(storage, 1)
```
CasperWA, line 72:
```suggestion
```

doc/transactions.md
```diff
@@ -0,0 +1,78 @@
+Transactions
+============
+Transactions is a feature that comes from SOFT, which allows to easy manage arbitrary long series of immutable (frozen) instances while ensuring provenance.  Conceptually it share many similarities with git.
```
CasperWA, line 3:
```suggestion
Transactions is a feature that comes from SOFT, which allows to easy manage arbitrary long series of immutable (frozen) instances, while ensuring provenance.
Conceptually, it shares many similarities with [git](https://git-scm.com).
```

doc/transactions.md
```diff
@@ -0,0 +1,78 @@
...
+
+A basic usage for transactions is to take snapshots of the current state of your system.  This is shown in Figure 1, where we assume that the instance `A` (which e.g. could be a collection) describes the state of your system.
```
CasperWA, line 5:
```suggestion
A basic usage for transactions is to take snapshots of the current state of your system.
This is shown in Figure 1, where we assume that the instance `A` (which, e.g., could be a collection) describes the state of your system.
```

doc/transactions.md
```diff
@@ -0,0 +1,78 @@
...
+
+We call an instance a *transaction* after we have taken a snapshot of it, i.e. when it contains a reference to a immutable "parent" instance.
```
CasperWA, line 7:
```suggestion
We call an instance a *transaction* after we have taken a snapshot of it, i.e. when it contains a reference to an immutable previous ("parent") instance.
```

doc/transactions.md
```diff
@@ -0,0 +1,78 @@
...
+
+![transactions](figs/transactions.png)
+
+**Figure 1**: *Creating a transaction by taking snapshots.  (a) initial state of your system described by instance `A`.  (b) creating a transaction by taking a snapshot
+of `A`.  The snapshot, `A1`, stores an immutable (frozen) copy of `A` from the exact
+moment `t1` the snapshot was taken.  (c) After another snapshot was taken at time `t2`.  Blue circles represent immutable instances, red circles represent mutable instances, while arrows relate an instance to its (frozen) parent instance.*
```
CasperWA, line 13:
```suggestion
**Figure 1**: *Creating a transaction by taking snapshots.
(a) initial state of your system described by instance `A`.
(b) creating a transaction by taking a snapshot of `A`.
The snapshot, `A1`, stores an immutable (frozen) copy of `A` from the exact moment `t1` the snapshot was taken.
(c) After another snapshot was taken at time `t2`.
Blue circles represent immutable instances, red circles represent mutable instances, while arrows relate an instance to its (frozen) parent instance.*
```

doc/transactions.md
```diff
@@ -0,0 +1,78 @@
...
+
+It is very simple to create such snapshots using the Python and C APIs.  In Python, the steps in Figure 1 can be produced by the following code:
```
CasperWA, line 15:
```suggestion
It is straight forward to create such snapshots using the Python and C APIs.
In Python, the steps in Figure 1 can be produced by the following code:
```

doc/transactions.md
```diff
@@ -0,0 +1,78 @@
...
+The corresponding code in C would be:
+
+```C
+DLiteInstance *A = dlite_instance_get("A");  // Initial state (Fig. 1a)
+dlite_instance_snapshot(A);                  // Make snapshot at time=t1 (Fig. 1b)
+// A is evolved...
+dlite_instance_snapshot(A);                  // Make snapshot at time=t2 (Fig. 1c)
+```
+
+The snapshots can be accessed using `dlite_instance_get_snapshot()`.  For instance, accessing snapshot `A2` and creating a new branch from it can be achieved by the following four lines of Python C code:
```
CasperWA, line 33:
```suggestion
The snapshots can be accessed using `dlite_instance_get_snapshot()`.
For instance, accessing snapshot `A2` and creating a new branch from it can be achieved by the following four lines of Python code:
```

doc/transactions.md
```diff
@@ -0,0 +1,78 @@
...
+
+```python
+A2 = A.get_snapshot()  # Access the most resent snapshot of A
```
CasperWA, line 36:
```suggestion
A2 = A.get_snapshot()  # Access the most recent snapshot of A
```

doc/transactions.md
```diff
@@ -0,0 +1,78 @@
...
+B = A2.copy()          # Create a mutable copy of A2
+B.set_parent(A2)       # Make B a transaction with A2 as parent. Shown Fig. 2a.
```
CasperWA, line 38:
```suggestion
B.set_parent(A2)       # Make B a transaction with A2 as parent. Shown in Fig. 2a.
```

doc/transactions.md
```diff
@@ -0,0 +1,78 @@
...
+B.shapshot()           # Make a snapshop of B. Shown Fig. 2b.
```
CasperWA, line 39:
```suggestion
B.shapshot()           # Make a snapshop of B. Shown in Fig. 2b.
```

doc/transactions.md
```diff
@@ -0,0 +1,78 @@
...
+```
+
+The corresponding C code is:
+
+```C
+// Access the most resent snapshot of instance A.
```
CasperWA, line 45:
```suggestion
// Access the most recent snapshot of instance A.
```

doc/transactions.md
```diff
@@ -0,0 +1,78 @@
...
+// Note that A2 is a borrowed reference to the snapshot and should not be dereferred
```
CasperWA, line 46: "dereferred" or "dereferenced"?

doc/transactions.md
```diff
@@ -0,0 +1,78 @@
...
+// with dlite_instance_decref().
+const DLiteInstance *A2 = dlite_instance_get_snapshot(A, 1);
+
+// Create a mutable copy of A2.
+// Note that B is a new reference.  You should call dlite_instance_decref() when you
+// are done with it.
+DLiteInstance *B = dlite_instance_copy(A2);
+
+// Make B a transaction with A2 as parent. Shown Fig. 2a.
```
CasperWA, line 55:
```suggestion
// Make B a transaction with A2 as parent. Shown in Fig. 2a.
```

doc/transactions.md
```diff
@@ -0,0 +1,78 @@
...
+dlite_instance_set_parent(B, A2);
+
+// Make a snapshop of B. Shown Fig. 2b.
```
CasperWA, line 58:
```suggestion
// Make a snapshop of B. Shown in Fig. 2b.
```

doc/transactions.md
```diff
@@ -0,0 +1,78 @@
...
+dlite_instance_snapshot(B);
+```
+The result if these commands are shown in Figure 2.
```
CasperWA, line 61:
```suggestion

The result if these commands are shown in Figure 2.
```

doc/transactions.md
```diff
@@ -0,0 +1,78 @@
...
+
+![transactions-branch](figs/transactions-branch.png)
+
+**Figure 2**: *Creating a new branch, from a snapshot of a transaction.  (a) create a copy `B` of latest snapshot of transaction `A`.  (b)  Take a snapshot of `B`.*
+
+
```
CasperWA, line 67:
```suggestion
```

doc/transactions.md
```diff
@@ -0,0 +1,78 @@
...
+dlite_instance_set_parent(B, A2);
+
+// Make a snapshop of B. Shown Fig. 2b.
+dlite_instance_snapshot(B);
+```
+The result if these commands are shown in Figure 2.
+
+![transactions-branch](figs/transactions-branch.png)
+
+**Figure 2**: *Creating a new branch, from a snapshot of a transaction.  (a) create a copy `B` of latest snapshot of transaction `A`.  (b)  Take a snapshot of `B`.*
```
CasperWA, line 65:
```suggestion
**Figure 2**: *Creating a new branch, from a snapshot of a transaction.
(a) Create a copy (`B`) of latest snapshot of transaction `A`.
(b) Take a snapshot of `B`.*
```

doc/transactions.md
```diff
@@ -0,0 +1,78 @@
...
+
+
+### Transaction parent and immutability
+All transactions starts with a root instance with no parent instance.  All other instances in a transaction has exactly one parent instance.
```
CasperWA, line 69:
```suggestion

All transactions start as a root instance with no parent instance.
All other instances in a transaction has exactly **one** parent instance.
```

doc/transactions.md
```diff
@@ -0,0 +1,78 @@
...
+All instances in a transaction that serves as a parent are immutable (that is, all instances except the leafs).  Non-root transaction instances store a [SHA-3](https://en.wikipedia.org/wiki/SHA-3) hash of their parents together with the parent UUIDs.  This make it possible to ensure that any of the ancestors of a transaction has not been changed - providing provenance.
```
CasperWA, line 70:
```suggestion
All instances in a transaction that serves as a parent are immutable (that is, all instances except the leaves).
Non-root transaction instances store a [SHA-3](https://en.wikipedia.org/wiki/SHA-3) hash of their parent together with the parent UUID.
This make it possible to ensure that any of the ancestors of a transaction has not been changed - providing provenance.
```

doc/transactions.md
```diff
@@ -0,0 +1,78 @@
...
+
+A transaction can be validated with the `verify_transaction()` method in Python and `dlite_instance_verify_transaction()` in C.
+
+
+### Memory management
```
CasperWA, line 75:
```suggestion
### Memory management

```

doc/transactions.md
```diff
@@ -0,0 +1,78 @@
...
+Since the number of snapshots potentially can be very large, it is important to be able to store them to disk in order to save memory.  To support this, DLite implements the `pull_snapshot()` and `push_snapshot()` methods (`dlite_instance_pull_snapshot()` and `dlite_instance_push_snapshot()` in C).
```
CasperWA, line 76:
```suggestion
The number of snapshots can potentially be very large, hence it is important to be able to store them to disk in order to save memory.
To support this, DLite implements the `pull_snapshot()` and `push_snapshot()` methods (`dlite_instance_pull_snapshot()` and `dlite_instance_push_snapshot()` in C).
```

src/CMakeLists.txt
```diff
@@ -40,6 +40,7 @@ set(sources
   dlite-storage-plugins.c
   dlite-mapping.c
   dlite-mapping-plugins.c
+  #dlite-transaction.c
```
CasperWA, line 43:
```suggestion
```

src/tests/test_entity.c
```diff
@@ -61,22 +61,22 @@ MU_TEST(test_meta_create)
   mu_assert_int_eq(1, dlite_instance_is_meta((DLiteInstance *)entity));
   mu_assert_int_eq(0, dlite_instance_is_metameta((DLiteInstance *)entity));
 
-  dlite_instance_debug((DLiteInstance *)entity);
+  //dlite_instance_debug((DLiteInstance *)entity);
```
CasperWA, line 65:
```suggestion
```

src/tests/test_entity.c
```diff
@@ -389,6 +389,139 @@ MU_TEST(test_instance_get_hash)
...
+  mu_assert_int_eq(0, dlite_instance_get_hash(snapshot1, h1, DLITE_HASH_SIZE));
+  mu_assert_int_eq(0, dlite_instance_get_hash(snapshot2, h2, DLITE_HASH_SIZE));
+
+  mu_assert_int_eq(1, inst->_refcount);
+  mu_assert_int_eq(2, snapshot1->_refcount);
+  mu_assert_int_eq(2, snapshot2->_refcount);
+
+  dlite_instance_decref(snapshot1);
+  dlite_instance_decref(snapshot2);
+  */
```
CasperWA, line 517: This section should be either removed or cleaned up.

CasperWA (changes requested):
Note, I have **not** reviewed any code/files related to C. Except a few commented out blocks that should be removed if not used.

bindings/python/tests/test_transaction.py
jesper-friis: Good point, we can add a few tests. But truly ensuring that the frozen flag is respected is not possible in C. All setter functions already checks for immutability, but if you have an array property, nothing prevents a user from getting a pointer to the array and change the memory it points to. 

In Python I have now added immutability checks to the `__setattr__()` and `__getattr__()` methods. `__getattr__()` checks if the returned object is a numpy array. If it is and the instance is frozen, the array is made read-only which will raise a ValueError if the user tries to assign to it.

bindings/python/tests/test_transaction.py
jesper-friis: Is such a for loop more readable than testing for explicit values?

bindings/python/tests/test_transaction.py
jesper-friis: This is a test to come when issue #261 has been implemented...

doc/transactions.md
jesper-friis: dereferenced is better

bindings/python/tests/test_transaction.py
CasperWA: Perhaps not - I suppose it's a matter of taste. Personally I like it most like this simply due to the fact that one can see the intended way the value is calculated - meaning the test as a whole becomes more transparent.

bindings/python/tests/test_transaction.py
CasperWA: I simply meant to add a test to try and change values in the snapshots and check an exception is raised :) But great!

bindings/python/dlite-entity-python.i
```diff
@@ -507,17 +512,23 @@ def get_instance(id: "str", metaid: "str"=None, check_storages: "bool"=True) ->
...
-            return _get_property(self, name)
+            value = _get_property(self, name)
         elif self.has_dimension(name):
-            return self.get_dimension_size(name)
+            value = self.get_dimension_size(name)
         else:
             raise AttributeError('Instance object has no attribute %r' % name)
+        if isinstance(value, np.ndarray) and self.is_frozen():
+            value.flags.writeable = False  # ensure immutability
+        return value
```
CasperWA, line 524: Again - I'm not sure why you need to consult `is_frozen()` here for _getting_ an attribute out of an instance.

bindings/python/dlite-entity-python.i
```diff
@@ -507,17 +512,23 @@ def get_instance(id: "str", metaid: "str"=None, check_storages: "bool"=True) ->
...
 
     def __setattr__(self, name, value):
         if name == 'this':
             object.__setattr__(self, name, value)
+        elif self.is_frozen():
+            raise DLiteError(
+                'frozen instance does not support attribute assignment')
```
CasperWA, line 531: Shouldn't this be the first check?

bindings/python/dlite-entity-python.i
```diff
@@ -481,15 +481,20 @@ def get_instance(id: "str", metaid: "str"=None, check_storages: "bool"=True) ->
...
         elif self.has_property(ind):
-            return self.get_property(ind)
+            value = self.get_property(ind)
         elif isinstance(ind, int):
             raise IndexError('instance property index out of range: %d' % ind)
         else:
             raise KeyError('no such property: %s' % ind)
+        if isinstance(value, np.ndarray) and self.is_frozen():
+            value.flags.writeable = False  # ensure immutability
+        return value
```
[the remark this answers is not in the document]
CasperWA: In the context of someone changing this array and it inadvertently also updates the instance's value due to memory-sharing, it makes sense - however, in the context of simply retrieving a value and manipulating it externally, it doesn't seem to make sense?

bindings/python/tests/test_transaction.py
CasperWA: This is not something you wanted to implement?

bindings/python/tests/test_transaction.py
```diff
@@ -0,0 +1,86 @@
...
+assert person.is_frozen() == False
+assert person.get_snapshot(1).is_frozen() == True
+assert person.get_snapshot(2).is_frozen() == True
+
+frozen = person.get_snapshot(1)
+assert frozen.is_frozen() == True
+try:
+    frozen.age = 77
+except dlite.DLiteError:
+    pass
```
CasperWA, line 46: You can even test the message in the exception here if you wish.

doc/transactions.md
```diff
@@ -0,0 +1,91 @@
+Transactions
+============
+Transactions is a feature that comes from SOFT, which allows to easy manage arbitrary long series of immutable (frozen) instances, while ensuring provenance.
```
CasperWA, line 3:
```suggestion
============

Transactions is a feature that comes from SOFT, which allows to easy manage arbitrary long series of immutable (frozen) instances, while ensuring provenance.
```

CasperWA: Do you have a reference to the part in SOFT where this comes from? If not, that's fine.

doc/transactions.md
```diff
@@ -0,0 +1,91 @@
...
+Conceptually, it shares many similarities with [git](https://git-scm.com).
+
+A basic usage for transactions is to take snapshots of the current state of your system.
+This is shown in Figure 1, where we assume that the instance `A` (which, e.g., could be a collection) describes the state of your system.
+
+We call an instance a *transaction* after we have taken a snapshot of it, i.e. when it contains a reference to an immutable previous ("parent") instance.
```
CasperWA, line 9: Is it possible to take a snapshot that is exactly similar in content to a previous one? This should be mentioned somewhere in this section.

In the figure example, e.g., it is unclear whether the _values_/content of `A1` has been changed, or the time `t1` is merely a meta-data value.
In general, the concept of time in these examples are a bit tenuous at best. It should be made more explicit - either by setting up a thought example about a time-resolved experiment, where time is a value and property of the underlying data model, or in another way that makes it clear what `t1` and `t2` are, other than abstract moments in time.

doc/transactions.md
```diff
@@ -0,0 +1,91 @@
...
+
+**Figure 2**: *Creating a new branch, from a snapshot of a transaction.
+(a) Create a copy (`B`) of latest snapshot of transaction `A`.
+(b) Take a snapshot of `B`.*
+
+### Transaction parent and immutability
+
+All transactions start as a root instance with no parent instance.
+All other instances in a transaction has exactly **one** parent instance.
+All instances in a transaction that serves as a parent are immutable (that is, all instances except the leaves).
```
CasperWA, line 80:
```suggestion
All instances in a transaction that serve as a parent are immutable (that is, all instances except the leaves (the latest)).
```

doc/transactions.md
```diff
@@ -0,0 +1,91 @@
...
+Non-root transaction instances store a [SHA-3](https://en.wikipedia.org/wiki/SHA-3) hash of their parent together with the parent UUID.
+This make it possible to ensure that any of the ancestors of a transaction has not been changed - providing provenance.
```
CasperWA, line 82:
```suggestion
This makes it possible to ensure that any of the ancestors of a transaction have not been changed - providing provenance.
```

doc/transactions.md
```diff
@@ -0,0 +1,91 @@
...
+
+A transaction can be validated with the `verify_transaction()` method in Python and `dlite_instance_verify_transaction()` in C.
```
CasperWA, line 84: Since the method names use "verify", it seemed more appropriate to use in the text as well, I think.
```suggestion
A transaction can be verified with the `verify_transaction()` method in Python and `dlite_instance_verify_transaction()` in C.
```

doc/transactions.md
```diff
@@ -0,0 +1,91 @@
...
+
+### Memory management
+
+The number of snapshots can potentially be very large, hence it is important to be able to store them to disk in order to save memory.
+To support this, DLite implements the `pull_snapshot()` and `push_snapshot()` methods (`dlite_instance_pull_snapshot()` and `dlite_instance_push_snapshot()` in C).
+
+Note that not all storages can be used with these functions, since whether an instance is a transactions or not, is not described by its metadata and hence requires special support by the storage plugin.
```
CasperWA, line 91: It might be nice to write a bit more about the future of memory management, specifically how it's stored on disk to save space and such? It doesn't need to be much, just some empty promises that similar to `git`, we might try to pack/bundle old snapshots together to save even more space on disk?

bindings/python/dlite-entity-python.i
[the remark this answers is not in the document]
jesper-friis: This make array properties of immutable instances read-only. The user can still still access and slice the array, but not change its values. Isn't that what we want?

bindings/python/dlite-entity-python.i
jesper-friis: Because the user can do
```python
v = frozen_inst.array_property  # calling __getattr__()
v[0, 0] = 42
```

CasperWA: Right. Hmm, there is a tricky bit of decision-making here. Because if one does a deep copy of the attribute, one can _still_ not change it, since the array values have been "frozen". However, this is not desireable.
On the other hand, we, of course, don't want a user to change _any_ values of a frozen instance...
Perhaps the middle ground here is to keep the arrays frozen, but then document this particular point in the documentation part, including some examples on how to manipulate the values of a frozen instance, if so desired (but not _actually_ changing the values within the instance).

CasperWA: Another question: Can a value be a `dict`? Actually, even if it can be a list, the container-specific methods might present a "danger" to the immutability of the instance.

For example, say I do the following:

```python
frozen_inst.list_value.append("test")
```

This shouldn't work, I suppose, but it might?

bindings/python/dlite-entity.i
```diff
@@ -382,6 +404,159 @@ Call signatures:
...
+  struct _DLiteInstance *pull_snapshot(struct _DLiteStorage *storage, int n) {
+    DLiteInstance *inst =
+      (DLiteInstance *)dlite_instance_pull_snapshot($self, storage, n);
+    if (inst) dlite_instance_incref(inst);
+    return inst;
+  }
+
+  %feature("docstring",
+           "Push all ancestors of snapshot `n` from memory to storage,\n"
+           "where `n=0` corresponds to `inst`, `n=1` to the parent of\n"
```
CasperWA, line 517:
```suggestion
           "where `n=0` corresponds to the instance, `n=1` to the parent of\n"
```

bindings/python/dlite-entity.i
```diff
@@ -382,6 +404,159 @@ Call signatures:
...
+           "`inst`, etc...\n"
```
CasperWA, line 518:
```suggestion
           "the instance, etc...\n"
```

bindings/python/dlite-entity.i
```diff
@@ -382,6 +404,159 @@ Call signatures:
...
+           "\n"
+           "This function starts pushing closest to the root to ensure\n"
+           "that the transaction is in a consistent state at all times."
+           ) push_snapshot;
+  void push_snapshot(struct _DLiteStorage *storage, int n) {
+    dlite_instance_push_snapshot($self, storage, n);
+  }
+
+  %feature("docstring",
+           "Turn instance `inst` into a transaction node with parent\n"
```
CasperWA, line 531:
```suggestion
           "Turn the instance into a transaction node with parent\n"
```

bindings/python/dlite-entity.i
```diff
@@ -382,6 +404,159 @@ Call signatures:
...
+           "`parent`.  This require that `inst` is mutable, and `parent`\n"
```
CasperWA, line 532:
```suggestion
           "`parent`.  This requires that the instance is mutable, and `parent`\n"
```

bindings/python/dlite-entity.i
```diff
@@ -382,6 +404,159 @@ Call signatures:
...
+           "is immutable.  If `inst` already has a parent, it will be\n"
```
CasperWA, line 533:
```suggestion
           "is immutable.  If the instance already has a parent, it will be\n"
```

bindings/python/dlite-entity.i
```diff
@@ -382,6 +404,159 @@ Call signatures:
...
+           "replaced.\n"
+           "\n"
+           "Use dlite_instance_freeze() and dlite_instance_is_frozen() to\n"
```
CasperWA, line 536:
```suggestion
           "Use `dlite_instance_freeze()` and `dlite_instance_is_frozen()` to\n"
```

bindings/python/dlite-entity.i
```diff
@@ -382,6 +404,159 @@ Call signatures:
...
+  }
+
+  %feature("docstring",
+           "Mark the instance as immutable.  This can never be reverted."
+           ) freeze;
+  void freeze(void) {
+    dlite_instance_freeze($self);
+  }
+
+  %feature("docstring", "Returns whether instance is immutable.") is_frozen;
```
CasperWA, line 550:
```suggestion
  %feature("docstring", "Returns whether the instance is frozen and immutable.") is_frozen;
```

doc/transactions.md
```diff
@@ -0,0 +1,95 @@
+Transactions
+============
```
CasperWA, line 2:
```suggestion
============

```

requirements.txt
```diff
@@ -3,6 +3,7 @@ numpy
 PyYAML
 psycopg2-binary
 pandas
+pint
 pymongo
 rdflib
 pint
```
CasperWA, line 9:
```suggestion
```
Already added above.

CasperWA (commented):
There are still several previous comments of mine you have not addressed.
Again, I cannot properly review the C code in this PR, and I wouldn't take an approval from me on this PR as a go-ahead for merging it until the C code has been properly reviewed as well.

doc/transactions.md
quaat: Just to be clear about nomenclature. When dealing with transactions we have to separate between identity, state and values. A value cannot change, 3 will always be 3, but the age of an item can change over time, but at any given time have a specific state. The state at a given point in time (specific instance of state/snapshot) is a value, and cannot be changed. identity is related to a sequence of states (snapshots/transactions) and point to different values (immutable) at a specific point in time (state).

quaat (approved):

bindings/python/dlite-entity-python.i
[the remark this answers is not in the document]
CasperWA: > I don't think we should return a copy, since we (deliberately) are not doing it in any other case. It is very easy for the user to make a copy of the returned value if he/she wants to freely manipulate it.

Well. When returning a string or a number (int/float), it's by definition always a copy in Python. This would just mimic this behaviour. But I see your point - letting the user do their thing.
Essentially, I just wanted a situation where whatever the user does, the values stored in the frozen DLite instance _cannot_ be changed. Including when returning memory pointers to those values to the user.

> Another reason for not making a copy is that it in most cases is just a waste of memory and CPU time. If the property is a 1000x1000x1000 array we really don't want any unnecessary duplication of memory.

Fair enough.

If this implementation ensures the optimal memory usage, while keeping the value completely immutable (through the standard Python API operations), then I'm very much happy :)

bindings/python/dlite-entity-python.i
CasperWA: > Dimension and Property properties are only intended to be used in metadata and never as scalars. I therefore don't think that they are a big issue.

As long as the user can't suddenly change the underlying data model in a weird way that's unintended, I suppose this is fine. Although even changing the descriptions of properties from instance to instance without a reference to an updated data model schema shouldn't be possible. Actually, that probably shouldn't be possible ever, without changing the version number of the reference data model schema.

> Relations are normally also used as arrays, but a user could potential define a scalar relation which will not be enforced to be non-writable in a frozen instance. We could add a new issue about it...

Good to an issue, but it perhaps doesn't seem like a thing users would do? But it's definitely an "exploitable", which should be "patched" at some point. So a follow-up issue on this is good.

bindings/python/dlite-entity-python.i
CasperWA: Hmm. We should add another issue about avoiding using proper keywords as special keywords, so users don't run into an issue with these special keywords suddenly.

I suggest to change the special keywords to something more straight-forward. E.g., `uuid` -> `_dlite_uuid` or something?

bindings/python/dlite-entity.i
```diff
@@ -382,6 +404,159 @@ Call signatures:
     dlite_instance_save(storage, $self);
   }
 
+  %feature("docstring", "Returns a hash of instance.") get_hash;
```
CasperWA, line 407:
```suggestion
  %feature("docstring", "Returns a hash of the instance.") get_hash;
```

or `an` if one does not pass in a specific instance to the function.

CasperWA (changes requested):
Just a single requested change - for completion.

Otherwise, I approve this PR, with the condition that issues are raised according to previous comment threads.

PR #270, opened 2022-06-02
Check for stream errors before calling assert

jesper-friis:


src/utils/jstore.c
```diff
@@ -32,6 +32,8 @@ char *jstore_readfp(FILE *fp)
 {
   char *q, *buf=NULL;
   size_t n, bytes_left, bytes_read=0, size=256;
+  if (feof(fp)) FAIL("stream error. Is it a regular file "
```
quaat, line 35: I suppose the regular file needs write-permission, not read-permission

quaat (changes requested):

src/utils/jstore.c
jesper-friis: The code below just reads from fp using fread()

jesper-friis: Updated the PR. Now it should catch pathetic FILE * objects and return a proper error

PR #276, opened 2022-07-10
Implemented a triplestore Python sub-package

### LINKED ISSUE #275: Triplestore Interface
Create a simple interface to different triplestores. Use the strategy design pattern for interfacing different RDF backends. Examples of possible backends that eventually could be supported are:
- list of tuples
- rdflib
- owlready2/EMMOntoPy
- Stardog
- DLite triplestore (based on Redland librdf)
- Redland librdf
- Apache Jena Fuseki
- Allegrograph

The idea is to model the interface after rdflib.Graph. The most important method to support are triples() and addN(). The convenient methods value(), subjects(), predicates(), objects(), subject_predicates(), subject_objects(), predicate_objects() and add() may also be nice to have. They can be implemented as a simple convenient layer on top of triples() and addN() (so no need for supporting them at the backend level). Like rdflib, these methods (except value(), add(), addN()) should return a generator to reduce memory consumption when dealing with large number of triples. 

Use the strategy design pattern when implementing the parse() and serialize() methods, to make it easy to combine and serialise to different triplestore backends. 

A possible simplification compared to rdflib, from a user perspective of view, could be to by default treat strings as IRIs. However, for the add() method, we would need a way to mark whether the object is an IRI or a literal. That could either be done by defining a `Literal` class like rdflib, or (maybe easier) to add an optional `type` and `lang` arguments to add(). If any of these are not None, the object is considered a literal, otherwise an IRI. For addN(), it might be easier with a `Literal` class.

query() and update() methods for SPARQL may also be nice to have. May not be supported by all backends.

Specialised methods should also be added for simplifying adding mappings and functions, like add_mapsTo() and add_function(). Note that add_mapsTo() may add a dependency on DLite. If that is unwanted, DLite could subclass the generic implementation and add these methods to the subclass. 

Example of a possible session
```python
def func(x):
    return np.mean(x)

ts = TripleStore(backend='rdflib')
ts.parse('onto.ttl')
ts.add_mapsTo('onto:my_ontological_concept', my_dlite_inst, 'my_property_name')
ts.add_function(func, expects=('onto:Prop1', 'onto:Prop2', ...), returns=('onto:Prop5', ))

print(f'subjects: ', [s for s, _, _ in ts.triples((None, ':my_predicate', ':my_object'))])
```

Convenient handling of namespace prefixes would also be a nice additional feature. Here we might deviate slightly from rdflib for simplicity and easy use. We could e.g. make the following possible:
```python
EMMO = ts.bind('emmo', 'http://emmo.info/emmo#')
ts.add_mapsTo(EMMO.Mass, 'http://onto-ns.com/meta/0.1/Molecule#molecule_weight')  # explicit alternative to providing DLite instance and property name
ts.namespaces  # => {'emmo': 'http://emmo.info/emmo#'}
```
Like rdflib, we could pre-populate the triplestore with a set of common prefixes and make them available as constants in the module
```python
from triplestore import TripleStore, RDF, RDFS, ...

RDFS.subClassOf  # => http://www.w3.org/2000/01/rdf-schema#subClassOf
```

jesper-friis:
Closes #275 

Further improvements, like those liste below should be separate issues. 

* Add additional backends. Candidates include:
    - Stardog
    - list of tuples
    - owlready2/EMMOntoPy
    - DLite triplestore (based on Redland librdf)
    - Redland librdf
    - Apache Jena Fuseki
    - Allegrograph
* Update the query() method to return the SPARQL result in a backend-independent way.
* Allow backends to be provided as plugins.

### DIFF
--- a/bindings/python/tests/test_triplestore.py
+++ b/bindings/python/tests/test_triplestore.py
@@ -1,0 +1,98 @@
+from dlite.triplestore import en, Literal, Triplestore, RDF, RDFS, XSD, OWL
+
+
+# Test namespaces
+assert str(RDF) == "http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+assert RDF.type == "http://www.w3.org/1999/02/22-rdf-syntax-ns#type"
+
+# Test RDF literal
+l1 = Literal("Hello world!")
+assert l1 == "Hello world!"
+assert isinstance(l1, str)
+assert l1.lang is None
+assert l1.datatype is None
+assert l1.toPython() == "Hello world!"
+assert l1.value == "Hello world!"
+assert l1.n3() == '"Hello world!"'
+
+l2 = Literal("Hello world!", lang="en")
+assert l2.lang == "en"
+assert l2.datatype == None
+assert l2.value == "Hello world!"
+assert l2.n3() == '"Hello world!"@en'
+
+l3 = en("Hello world!")
+assert l3.n3() == '"Hello world!"@en'
+
+l4 = Literal(42)
+assert l4.lang == None
+assert l4.datatype == XSD.integer
+assert l4.value == 42
+assert l4.n3() == f'"42"^^{XSD.integer}'
+
+l5 = Literal(42, datatype=float)
+assert l5.lang == None
+assert l5.datatype == XSD.double
+assert l5.value == 42.0
+assert l5.n3() == f'"42"^^{XSD.double}'
+
+# Test rdflib triplestore backend
+ts = Triplestore("rdflib")
+assert ts.expand_iri("xsd:integer") == XSD.integer
+assert ts.prefix_iri(RDF.type) == 'rdf:type'
+EX = ts.bind("ex", "http://example.com/onto#")
+assert str(EX) == "http://example.com/onto#"
+ts.add_mapsTo(EX.MyConcept, "http://onto-ns.com/meta/0.1/MyEntity", "myprop")
+ts.add((EX.MyConcept, RDFS.subClassOf, OWL.Thing))
+ts.add((EX.AnotherConcept, RDFS.subClassOf, OWL.Thing))
+ts.add((EX.Sum, RDFS.subClassOf, OWL.Thing))
+
+def sum(a, b):
+    """Summarise `a` and `b`."""
+    return a + b
+
+ts.add_function(sum, expects=(EX.MyConcept, EX.AnotherConcept), returns=EX.Sum,
+                base_iri=EX)
+
+s = ts.serialize(format="turtle")
+fid = id(sum)  # sum() function id
+assert s == f"""\
+@prefix dcterms: <http://purl.org/dc/terms/> .
+@prefix ex: <http://example.com/onto#> .
+@prefix fno: <https://w3id.org/function/ontology#> .
+@prefix map: <http://emmo.info/domain-mappings#> .
+@prefix owl: <http://www.w3.org/2002/07/owl#> .
+@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
+@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
+
+ex:sum_{fid} a fno:Function ;
+    dcterms:description "Summarise `a` and `b`."@en ;
+    fno:expects ( ex:sum_{fid}_parameter1_a ex:sum_{fid}_parameter2_b ) ;
+    fno:returns ( ex:sum_{fid}_output1 ) .
+
+<http://onto-ns.com/meta/0.1/MyEntity#myprop> map:mapsTo ex:MyConcept .
+
+ex:AnotherConcept rdfs:subClassOf owl:Thing .
+
+ex:Sum rdfs:subClassOf owl:Thing .
+
+ex:sum_{fid}_output1 a fno:Output ;
+    map:mapsTo ex:Sum .
+
+ex:sum_{fid}_parameter1_a a fno:Parameter ;
+    rdfs:label "a"@en ;
+    map:mapsTo ex:MyConcept .
+
+ex:sum_{fid}_parameter2_b a fno:Parameter ;
+    rdfs:label "b"@en ;
+    map:mapsTo ex:AnotherConcept .
+
+ex:MyConcept rdfs:subClassOf owl:Thing .
+
+"""
+
+ts2 = Triplestore("rdflib")
+ts2.parse(format="turtle", data=s)
+assert ts2.serialize(format="turtle") == s
+ts2.set((EX.AnotherConcept, RDFS.subClassOf, EX.MyConcept))
+# print(ts2.serialize(format="turtle"))
--- a/bindings/python/tests/test_triplestore.py
+++ b/bindings/python/tests/test_triplestore.py
@@ -1,17 +1,21 @@
 from dlite.triplestore import en, Literal, Triplestore, RDF, RDFS, XSD, OWL
+from dlite.triplestore.triplestore import function_id
 
 
 # Test namespaces
+# ---------------
 assert str(RDF) == "http://www.w3.org/1999/02/22-rdf-syntax-ns#"
 assert RDF.type == "http://www.w3.org/1999/02/22-rdf-syntax-ns#type"
 
-# Test RDF literal
+
+# Test RDF literals
+# -----------------
 l1 = Literal("Hello world!")
 assert l1 == "Hello world!"
 assert isinstance(l1, str)
 assert l1.lang is None
 assert l1.datatype is None
-assert l1.toPython() == "Hello world!"
+assert l1.to_python() == "Hello world!"
 assert l1.value == "Hello world!"
 assert l1.n3() == '"Hello world!"'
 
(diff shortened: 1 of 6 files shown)


bindings/python/tests/test_triplestore.py
```diff
@@ -0,0 +1,98 @@
+from dlite.triplestore import en, Literal, Triplestore, RDF, RDFS, XSD, OWL
+
+
+# Test namespaces
+assert str(RDF) == "http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+assert RDF.type == "http://www.w3.org/1999/02/22-rdf-syntax-ns#type"
+
+# Test RDF literal
```
daniel-sintef, line 8: Many of these variables are not very descriptive, consider renaming. E.g. l1 = literal_helloworld

daniel-sintef: The comment string is not very helpful 'testing literals' can be guessed by context. Consider adding more description. There are many different cases: l1,l2,l3,... consider adding comments as to why we are testing all these cases and what they are

bindings/python/tests/test_triplestore.py
```diff
@@ -0,0 +1,98 @@
...
+assert l4.value == 42
+assert l4.n3() == f'"42"^^{XSD.integer}'
+
+l5 = Literal(42, datatype=float)
+assert l5.lang == None
+assert l5.datatype == XSD.double
+assert l5.value == 42.0
+assert l5.n3() == f'"42"^^{XSD.double}'
+
+# Test rdflib triplestore backend
```
daniel-sintef, line 39: Again, this comment does not give us much information. What are we testing in the backend, how are we testing it, why are we testing in the way that we are?

bindings/python/triplestore/backends/rdflib.py
```diff
@@ -0,0 +1,108 @@
...
+
+if TYPE_CHECKING:  # pragma: no cover
+    from collections.abc import Sequence
+    from typing import Generator
+
+    from dlite.triplestore import Triple
+
+
+def asuri(v):
+    """Help function converting a spo-value to proper rdflib type."""
```
daniel-sintef, line 17: What is an 'spo-value' what is a 'proper' (as opposed to 'improper') rdflib type?

bindings/python/triplestore/backends/rdflib.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def astriple(t):
+    """Help function converting a triple to rdflib triple."""
+    s, p, o = t
+    return asuri(s), asuri(p), asuri(o)
+
+
+class RdflibStrategy:
+    """Triplestore strategy for rdflib."""
```
daniel-sintef, line 34: Maybe we should have some examples of usage in the docstring?

bindings/python/triplestore/interface.py
```diff
@@ -0,0 +1,33 @@
...
+from typing import TYPE_CHECKING, Protocol
+
+if TYPE_CHECKING:  # pragma: no cover
+    from collections.abc import Sequence
+    from typing import Generator
+
+    from dlite.triplestore import Triple
+
+
+class ITriplestore(Protocol):
```
daniel-sintef, line 11: Consider adding an example using this class

bindings/python/triplestore/triplestore.py
```diff
@@ -0,0 +1,608 @@
...
+
+    >>> def mean(x, y):
+    ...     """Returns the mean value of `x` and `y`."""
+    ...     return (x + y)/2
+
+    >>> ts.add_function(mean, expects=(ONTO.RightArmLength, ONTO.LeftArmLength),
+    ...                 returns=ONTO.AverageArmLength)
+
+
+TODO:
```
daniel-sintef, line 86: Probably should turn these TODO's into issues and remove them from the code

bindings/python/triplestore/triplestore.py
```diff
@@ -0,0 +1,608 @@
+'''A module encapsulating different triplestores using the strategy design
+pattern.
```
daniel-sintef, line 2: The level of self-documentation seems uneven. Here there is a lot of information at the top of the file, but interface.py has none...

bindings/python/triplestore/triplestore.py
```diff
@@ -0,0 +1,608 @@
...
+
+This module has no dependencies outside the standard library, but the
+triplestore backends have.
```
daniel-sintef, line 5: Is this critical to mention here? Probably should be moved to a top-level README?

bindings/python/triplestore/triplestore.py
```diff
@@ -0,0 +1,608 @@
...
+        lang = f", lang='{self.lang}'" if self.lang else ""
+        datatype = f", datatype='{self.datatype}'" if self.datatype else ""
+        return f"Literal('{self}'{lang}{datatype})"
+
+    value = property(
+        fget=lambda self: self.toPython(),
+        doc="Appropriate python datatype derived from this RDF literal.",
+    )
+
+    def toPython(self):
```
daniel-sintef, line 218: CamelCase is typically reserved for classes, this should probably be to_python

daniel-sintef (approved):

bindings/python/tests/test_triplestore.py
jesper-friis: Thanks for the review. 

This is just a test script, where the comments are more ment for grouping the different types of tests rather than describing them. I underlined them to make that more clear. Do you feel more documentation is needed after reading the documentation of the triplestore module?

Regarding variable names there exists many opinions. My preferences are based on the following considerations:
- variable names should be concise and easy understandable from their context
- the name should describes what the variable is, rather than how it is used
- short code lines are easier to read than long code lines (especially if you have a tendency to dyslexia) - hense avoid unnecessary long variable names
- don't reuse the same variable name for different things

Hence, I don't think that `l1` is necessary a bad variable name, since its meaning is rather clear from the context. It is only used in a short code section, clearly separated by blank lines, where the first line
```python
l1 = Literal("Hello world!")
```
makes it easy to see that it is a literal. An alternative name could be `literal1`, but I think that `l` is a recognisable abbreviation of "literal" given the context. The number `1` is added because we are instantiating several literals and don't want to use the same name for them. It also emphasises that `l1` is just an instance.

Personally I think a name like `literal_helloworld` is to long and just make the code harder to read. What is interesting is that we are testing a literal, not that its value is "Hello world!". 

I think this is a very good issue to discuss and comment on, since people have different preferences and there are no absolute true or false, except for that consistency counts! Hence, we should try to follow the same practice within the team...

bindings/python/triplestore/backends/rdflib.py
jesper-friis: It is an abbreviation for subject-predicate-object (i.e. an RDF triple). "spo" is common jargon in the field.

bindings/python/triplestore/backends/rdflib.py
jesper-friis: This class implements the rdflib strategy (backend) and is not intended to be used directly by the users. 

I think it is better to put the usage examples in the module documentation of triplestore.py, which provide the "public" interface exposed to the users. The triplestore module has already some usage examples. Is there something you miss?

bindings/python/triplestore/interface.py
jesper-friis: I have no clue of how to use it (except for inspection). ITriplestore is just a protocol class that formally define the interface that should be implemented by the. I added som additional documentation explaining this.

bindings/python/triplestore/triplestore.py
jesper-friis: Agree. Should be done...

bindings/python/triplestore/triplestore.py
jesper-friis: That is because this file provides the interface that the users are expected to see. The other files are internal to the package and are therefore much less documented.

But we could consider to move the module docstring from triplestore.py to the main __init__.py file if you think that would make more sense.

bindings/python/triplestore/triplestore.py
jesper-friis: Agree, it does not really belong together with the rest of the module documentation. But I still think it is a useful comment for the developer. 

If we move the module docstring to the `__init__.py` file, we could leave this comment where it is. Otherwise we could make it a separate comment. Placing it in the README file is also a good option...

bindings/python/triplestore/triplestore.py
jesper-friis: Good point. The reason for the CamelCase was to be consistent with rdflib. But I agree with you that internal consistency is more important than consistency with rdflib.

The function is now renamed to `to_python()`.

bindings/python/triplestore/triplestore.py
```diff
@@ -0,0 +1,608 @@
...
+            ns = namespace if isinstance(namespace, Namespace) else Namespace(
+                namespace)
+            self.namespaces[prefix] = ns
+
+        if hasattr(self.backend, "bind"):
+            self.backend.bind(prefix, namespace)
+
+        return ns
+
+    # --- Convenient methods
```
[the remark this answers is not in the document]
jesper-friis: This comment was meant as a sectioning header to make it easier for the developer to distinguish different types of methods implemented by the `Triplestore` class.

I underlined the comment to make its purpose more clear. Is it more helpful now?

Issue #222, opened 2022-02-28
Index plugins


CasperWA:
Plugins should be indexed in a separate text file to facilitate proper look-ups for DLite.

Currently DLite searches through all plugin files within the registered plugin directories resulting in import errors if plugins depend on packages that are not installed.
This is the case, even if a particular "problematic" plugin is not the intended target for a certain operation.

Suggestion:
Have a dedicated index file, e.g., a YAML file, in the root of each plugin folder that DLite can visit to look up the plugins available in the current folder.

jesper-friis:
Currently "problematic plugins" are ignored with a warning. 

I am not sure if it is more user friendly to add a text file listing available plugins that the user will have to maintain?

PR #293, opened 2022-08-04
Recover from invalid storage

### LINKED ISSUE #292: Recover from invalid storage in storage path
If a file in DLITE_STORAGES (dlite.storage_path in Python) has a unknown file extension and cannot be loaded, the search stops with an error instead of continue to see if there is a storage that can be read and contain the instance asked for.

jesper-friis:
Closes #292 

**Previous behaviour**:
If there was an invalid storage in the DLITE_STORAGES environment variable (or dlite.storage_path Python variable), then dlite would fail with an error message. This means that if there for some reason was an invalid storage (e.g. a non-dlite json configuration file) in the storage search path, dlite would fail immediately instead of continue the search for a valid storage.

**New behaviour**:
DLite will silently skip invalid storages in the storage search path and continue the search for a valid storage. Hence, instead of failing, dlite.get_instance() will return the wanted instance instead of failing. The user is also spared from seeing a lot of confusing (and most often harmless) error messages.

**Note**: 
The purpose of the storage search path is to allow to register one or some selected storages to act as a data space allowing to access metadata (and other instances) by id by calling dlite.get_instance().

This fix may also remove the motivation for issue #222


daniel-sintef:
Silent skipping can be a bit risky. I am a bit confused when you say "often" harmless, this would imply that it is sometimes harmful, and now it will not give any warning/error? Perhaps there can be a 'verbose' flag set up (which may be useful again later) which can be set, turning this warning on/off?

jesper-friis:
> Silent skipping can be a bit risky. I am a bit confused when you say "often" harmless, this would imply that it is sometimes harmful, and now it will not give any warning/error? Perhaps there can be a 'verbose' flag set up (which may be useful again later) which can be set, turning this warning on/off?

Yes, silently ignoring errors is normally bad (and maybe also in this case). But remember that we are speaking about dlite_instance_get() - a function that is provided the UUID of an instance and (magically) will return the corresponding instance, fully instantiated. The function is not told where to look for the instance, so it does the following heuristics:
1. Check if the instance is already loaded (exists in the internal in-memory storage). If so, return it.
2. Otherwise check if the instance exists in any hotlisted storage (an storage that is already opened). If so, return it.
3. Otherwise open all storages listed in the DLITE_STORAGE environment variable (or dlite.storage_path in Python) one by one and check for the instance. If found, it is returned.

The issue is that dlite will produce an error message for each storage it checks that doesn't contain an instance with the given UUID. If the user have many hot storages or many storages listed in DLITE_STORAGES, the screen will be cluttered with a lot of irrelevant error messages. 

This PR ignores all the messages originating from the underlying storages. If the instance is found, it is returned with no errors reported otherwise a single error is produced.

Do you have a suggestion for a more proper behaviour?

There is one problem with this though that require some work to handle properly. There may be several reasons why an instance cannot be loaded from a storage, like: 
- the instance isn't in the storage - this error we want to ignore
- the storage doesn't exists - this error we probably want to ignore
- the instance is in the storage, but we are not able to find its metadata - this error we would like to see
- the storage exists but is corrupted or invalid - this error we probably would like to see

I think the best and right way to solve this is to give each of these types of errors a separate error code and only ignoring those we want to ignore. DLite already supports error codes, but has been very sloppy using them (most functions simply returns zero on success and 1 (or -1) on failure).

src/dlite-entity.c
```diff
@@ -716,7 +716,13 @@ DLiteInstance *dlite_instance_get(const char *id)
 
     /* Set read-only as default mode (all drivers should support this) */
     if (!options) options = "mode=r";
-    if ((s = dlite_storage_open(driver, location, options))) {
+    ErrTry:
+      s = dlite_storage_open(driver, location, options);
+    ErrCatch(dliteStorageOpenError):  // suppressed error
```
daniel-sintef, line 721: As discussed some way of tracking/logging these errors would be useful

src/dlite-errors.h
```diff
@@ -14,10 +14,10 @@ typedef enum {
   dliteNullReferenceError,
   dliteStorageOpenError,     /*!< cannot open storage plugin */
   dliteStorageLoadError,     /*!< cannot load storage plugin */
+  dliteStorageSaveError,     /*!< cannot save storage plugin */
```
daniel-sintef, line 17: Why did this not exist before?

src/dlite-json.c
```diff
@@ -1076,7 +1076,7 @@ DLiteJsonFormat dlite_jstore_loads(JStore *js, const char *src, int len)
   unsigned int ntokens=0;
   char uuid[DLITE_UUID_LENGTH+1], *uri=NULL;
   int r;
-  DLiteJsonFormat format=-1;
+  DLiteJsonFormat retval=-1, format;
```
daniel-sintef, line 1079: Not very clear to me what this new 'format' variable is, why it has been added or what it does.

src/dlite-json.c
```diff
@@ -1116,11 +1116,12 @@ DLiteJsonFormat dlite_jstore_loads(JStore *js, const char *src, int len)
       t += jsmn_count(v) + 2;
     }
   }
+  retval = format;
 
  fail:
   if (tokens) free(tokens);
   if (uri) free(uri);
-  return format;
+  return retval;
```
daniel-sintef, line 1124: A bit confusing why retval = format? Seems like a confusing naming convention. Why not just return the format??

src/dlite-json.h
```diff
@@ -25,8 +25,9 @@ typedef enum {
 
 /** Enum indicating whether a JSON string is formatted as data and metadata. */
 typedef enum {
-  dliteJsonDataFormat,    /*!< Data format */
-  dliteJsonMetaFormat     /*!< Metadata format */
+  dliteJsonUnknownFormat=-1,  /*!< Unknown format */
+  dliteJsonDataFormat=0,      /*!< Data format */
```
daniel-sintef, line 29: DataFormat used to just be declared 'as-is' why do we set it to 0 now?

src/dlite-storage.c
```diff
@@ -88,8 +88,8 @@ DLiteStorage *dlite_storage_open(const char *driver, const char *location,
   if (!(api = dlite_storage_plugin_get(driver))) goto fail;
   if (!(s = api->open(api, location, options))) goto fail;
   s->api = api;
-  if (!(s->location = strdup(location))) FAIL(NULL);
-  if (options && !(s->options = strdup(options))) FAIL(NULL);
```
daniel-sintef, line 92: Difference between NULL and allocation failure?

src/utils/err.h
```diff
@@ -621,7 +621,7 @@ ErrRecord *_err_get_record();
     }                                                \
     _ft = _fallthrough;                              \
     _fallthrough = 0;                                \
-    _last = _record.eval == errval || _ft;           \
+    _last = (_record.eval == errval) || _ft;         \
```
daniel-sintef, line 624: Just curious what does the parenthesis do in this context? (maybe just  a basic c-programming question...)

src/dlite-entity.c
jesper-friis: Added issue #328 to address your previous comment. I guess that will address this one as well.

src/dlite-errors.h
jesper-friis: It wasn't used before, but it will be in #328.

src/dlite-json.c
jesper-friis: The function returns the json `format` or -1 on error.  Before there was a possibility that this function returned a json format, despite there was an error. This bug was corrected by introducing the `retval` variable which is initialised to -1 and only set to a (non-negative) format value if the function succeeds.

src/dlite-json.c
jesper-friis: See above

src/dlite-json.h
jesper-friis: This is an enum. Normally you will the compiler assigning the values instead of setting them explicit yourselves. 

However, it turns out that the compiler (at least gcc 11.3.1) by default will make DLiteJsonFormat an unsigned integer, which will mess up the negative return value of dlite_jstore_loads(). By explicitely setting `liteJsonUnknownFormat=-1` we force the compiler to make the enum a signed integer.

src/dlite-storage.c
jesper-friis: "allocation failure" is more explicit. Updated

src/utils/err.h
jesper-friis: They are not needed, but makes the precedence more explicit. 

While debugging, I checked whether the issue could be due to the precedence between `==` and `||` operators. Added the parenthesis to avoid being unsure again

daniel-sintef (approved):

PR #300, opened 2022-08-08
WIP: Support labels in dlite.triplestore

jesper-friis:


bindings/python/triplestore/triplestore.py
```diff
@@ -36,23 +38,154 @@ class UniquenessError(TriplestoreError):
...
+                if self._cache is not None:
+                    self._cache[name] = iri
+                return iri
+
+            # Check for label annotations matching `name`.
+            for la in self._label_annotations:
+                for s, o in self._triplestore.subject_objects(la):
+                    if name == o:
+                        if self._cache is not None:
+                            self._cache[name]: s
```
quaat, line 170: Did you mean to declare an assignment?

` self._cache[name] = s`

bindings/python/triplestore/triplestore.py
```diff
@@ -69,9 +202,9 @@ def __str__(self):
 PROV = Namespace("http://www.w3.org/ns/prov#")
 DCAT = Namespace("http://www.w3.org/ns/dcat#")
 TIME = Namespace("http://www.w3.org/2006/time#")
-GEO = Namespace("http://www.w3.org/2003/01/geo/wgs84_pos#")
-SDO = Namespace("http://schema.org/")
-VANN = Namespace("http://purl.org/vocab/vann/")
+#GEO = Namespace("http://www.w3.org/2003/01/geo/wgs84_pos#")
```
quaat, line 205: Remove lines

bindings/python/tests/test_triplestore.py
```diff
@@ -1,19 +1,71 @@
...
+from pathlib import Path
+
 try:
     import rdflib
 except ImportError:
     import sys
-    sys.exit(44)
+    sys.exit(44)  # skip this test
+
+try:
```
quaat, line 10: What is the advantage of this construction rather that just requiring pytest to be available upon running tests?

quaat (commented):
Great test-setup, just minor comments/questions

bindings/python/triplestore/triplestore.py
jesper-friis: Yes, that was the intention. Well spottet! 

I see now that there is another issue with this code as well. We only want to cache and return `s` if it belongs to the current namespace. Hence, I updated line 168 to
```python
    if o == name and s.startswith(self._iri):
```

bindings/python/triplestore/triplestore.py
jesper-friis: done

bindings/python/tests/test_triplestore.py
jesper-friis: Only to avoid a hard dependency on pytest. But I am not sure if it best practices though...

Writing the test as a script instead as a function as pytest requires, gives me the full local scope under my hands if during development by running the test via `ipython -i test_triplestore.py`. That is very convenient for testing different solutions and checking documentation.

Hence, I do not use pytest in general, except for pytest.raises(), which I find very nice and convenient for testing exceptions.

quaat (approved):

PR #307, opened 2022-08-09
Fix mapping costs

jesper-friis:
Corrected calculation of mapping costs.

TODO: check MappingStep.eval() including MappingStep.get_inputs()


bindings/python/mappings.py
```diff
@@ -228,14 +229,25 @@ def lowest_costs(self, nresults=5):
...
         n = 0
         for inputs in self.input_routes:
-            owncost = 1
-            for cost, idx in get_lowest_costs(inputs, nresults=nresults):
-                if isinstance(self.cost, Callable):
-                    values = get_values(inputs, idx, magnitudes=True)
-                    owncost = self.cost(**values)
+            costs = np.array([0.0])
+            for input in inputs.values():
+                if isinstance(input, MappingStep):
```
francescalb, line 234: Can you add a comment on how costs are calculated for MappingStep?

bindings/python/mappings.py
```diff
@@ -228,14 +229,25 @@ def lowest_costs(self, nresults=5):
...
+                    costs = np.tile(costs, len(res)) + np.repeat(
+                        [c for c, _ in sorted(res, key=lambda x: x[1])],
+                        len(costs))
                 else:
-                    owncost = self.cost
-                result.append((cost + owncost, n + idx))
+                    costs += input.cost
+
+            res = sorted((c, i) for i, c in enumerate(costs))[:nresults]
+            if isinstance(self.cost, Callable):
```
francescalb, line 243: Can you add a comment on how costs are calculated for Callable and how this differs from MappingStep?

bindings/python/mappings.py
jesper-friis: Done

bindings/python/mappings.py
jesper-friis: Done

francescalb (approved):
Approved under condition that an issue is made about a more thorough description of the algorithm and thoughts behind :)

PR #311, opened 2022-08-11
Interface for adding mapping cost 

jesper-friis:
Added options to add_mapsTo() and add_function() to provide a user-defined cost. 

The cost may be a float or a function returning the cost. Note that even if the cost is specified as a function, it is only represented in the knowledgebase as a literal identifier, without a fully ontological description of its arguments using e.g. FnO.


francescalb (approved):
Please make a test of the new  functions.

PR #312, opened 2022-08-11
Improved Python interface for dlite properties

jesper-friis:
* Added convinient dimnames() and propnames() functions to Metadata.
* Changed some occurences of `dims` to `shape` in Python interface to properties
* Allow to set property shape
* Made Property.ndims immutable
* Added tests

### DIFF
--- a/bindings/python/dlite-entity-python.i
+++ b/bindings/python/dlite-entity-python.i
@@ -193,9 +193,16 @@ def get_instance(id: "str", metaid: "str"=None, check_storages: "bool"=True) -> 
 
     type = property(get_type, doc='Type name.')
     dtype = property(get_dtype, doc='Type number.')
-    dims = property(get_dims, doc='Array of dimension indices. '
-                    '`dims` is deprecated, use `shape` instead.')
-    shape = property(get_dims, doc='Array of dimension indices.')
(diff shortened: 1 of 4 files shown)


bindings/python/dlite-entity-python.i
```diff
@@ -67,6 +67,14 @@ class Metadata(Instance):
             return lst[0]
         raise DLiteError(f"Metadata {self.uri} has no such property: {name}")
 
+    def dimnames(self):
+        """Returns a list of all dimension names in this metadata."""
+        return [d.name for d in self.properties['dimensions']]
```
quaat, line 72: Throw an informative error in the unlikely event that properties does not contain 'dimensions'

bindings/python/dlite-entity-python.i
```diff
@@ -67,6 +67,14 @@ class Metadata(Instance):
...
+
+    def propnames(self):
+        """Returns a list of all property names in this metadata."""
+        return [p.name for p in self.properties['properties']]
```
quaat, line 76: Throw an informative error in the unlikely event that properties does not contain 'properties'

quaat (commented):
Minor comments.

Issue #80, opened 2021-04-19
Simplify the storage plugin API


jesper-friis:
The storage plugin API defined in the bottom of src/dlite-storage-plugins.h has grown unnecessary large. 

- Implement full support CRUD, that is, add `delete()`
- getUUIDs can probably be removed and replaced with the iterators for queuing.
- Deprecate the old datamodel API since instances themselves fully support the data model. Hence, in the plugin, you can work with the getters and setters directly on the instances instead via the separate datamodel.
- `getDataName() / `setDataName()` can definitely be removed. The URI fulfils the role of a human readable name for instances and metadata.
- Consider to add optional `getData()` / `setData()` methods for writing/reading data to/from a memory buffer instead of a file storage. 
- Consider to add an optional `flush()` or `commit()` method that stores the data to the storage without closing it. 
- Consider separate schema from driver (or strategy) as done in OTEAPI.

PR #315, opened 2022-08-17
MongoDB plugin

### LINKED ISSUE #318: Fix iterate over instances/uuids in storage
`storage.instances()` and `storage.get_uuids()` works. But calling these functions with a metadata uri for filtering the result results in an error - at least for the mongodb backend...

jesper-friis:
Added initial stub for a MongoDB plugin written in python.

Also fixed an old bug in the C code for querying python storages (closes #318).


requirements_dev.txt
```diff
@@ -1,3 +1,4 @@
 ipython==7.31.1
 ipykernel==6.0.1
 ipython_genutils==0.2.0
+mongomock
```
CasperWA, line 4: If the other dependencies have a fixed version, this would also be good to have it, I suppose?
```suggestion
mongomock==4.1.2
```

storages/python/python-storage-plugins/mongodb.py
```diff
@@ -0,0 +1,94 @@
...
+        - mode: "r" for opening in read-only mode, "w" for read/write mode.
+        - authMechanism: Authentication mechanism (default: "SCRAM-SHA-256")
+        - mock: Whether to use mongomock.
+        """
+        self.options = Options(
+            options,
+            defaults='database=dlite;collection=dlite_coll;mode=w;'
+            'authMechanism=SCRAM-SHA-256;mock=no',
+        )
+        opts = self.options
```
CasperWA, line 39: Is there a reason you are redefining the `self.options` attribute locally here? It seems you could do without this and simply use `self.options` anywhere you are currently using the less clear `opts` variable?

storages/python/python-storage-plugins/mongodb.py
```diff
@@ -0,0 +1,94 @@
...
+                username=user,
+                password=password,
+                authSource=opts.database,
+                authMechanism=opts.authMechanism,
+            )
+            return client
+
+        self.client = get_client()
+        self.collection = self.client.get_database(
+            opts.database).get_collection(opts.collection)
```
CasperWA, line 71: See the [pymongo docs](https://pymongo.readthedocs.io/en/stable/tutorial.html#getting-a-database).
```suggestion
        self.collection = self.client[opts.database][opts.collection]
```

storages/python/python-storage-plugins/mongodb.py
```diff
@@ -0,0 +1,94 @@
...
+    def save(self, inst):
+        """Stores `inst` in current storage."""
+        document = inst.asdict(uuid=True)
+        self.collection.insert_one(document)
+
+    def queue(self, pattern=None):
+        """Generator method that iterates over all UUIDs in the storage
+        who's metadata URI matches glob pattern `pattern`."""
+        d = {"meta": pattern} if pattern else {}
+        for doc in self.collection.find(d):
```
CasperWA, line 93: Adding `projection` will reduce traffic and only return the "uuid" (and "_id").
```suggestion
        mongo_filter = {"meta": pattern} if pattern else {}
        for doc in self.collection.find(filter=mongo_filter, projection=["uuid"]):
```

storages/python/tests-python/test_mongodb_python.py
```diff
@@ -0,0 +1,54 @@
...
+
+# Load existing test data
+meta = dlite.Instance.from_location("json", inputdir / "test_meta.json")
+inst1 = dlite.Instance.from_location("json", inputdir / "test_data.json",
+                                     id="52522ba5-6bfe-4a64-992d-e9ec4080fbac")
+inst2 = dlite.Instance.from_location("json", inputdir / "test_data.json",
+                                     id="2f8ba28c-add6-5718-a03c-ea46961d6ca7")
+
+# Create storage
+s = dlite.Storage(
```
CasperWA, line 24:
```suggestion
storage = dlite.Storage(
```

storages/python/tests-python/test_mongodb_python.py
```diff
@@ -0,0 +1,54 @@
...
+    "mongodb", "localhost", "user=testuser;password=testpw;mock=true"
+)
+
+# Store metadata and data
+meta.save(s)
+inst1.save(s)
+inst2.save(s)
```
CasperWA, line 31:
```suggestion
meta.save(storage)
inst1.save(storage)
inst2.save(storage)
```

storages/python/tests-python/test_mongodb_python.py
```diff
@@ -0,0 +1,54 @@
...
+uuid1 = inst1.uuid
+uuid2 = inst2.uuid
+del inst1
+del inst2
+del meta
+
+# Load metadata and data
+new_meta = s.load(id=uuid0)
+new_inst1 = s.load(id=uuid1)
+new_inst2 = s.load(id=uuid2)
```
CasperWA, line 46:
```suggestion
new_meta = storage.load(id=uuid0)
new_inst1 = storage.load(id=uuid1)
new_inst2 = storage.load(id=uuid2)
```

storages/python/tests-python/test_mongodb_python.py
```diff
@@ -0,0 +1,54 @@
...
+
+# Simple queries
+instances = list(s.instances())
```
CasperWA, line 49:
```suggestion
instances = list(storage.instances())
```

storages/python/tests-python/test_mongodb_python.py
```diff
@@ -0,0 +1,54 @@
...
+assert len(instances) == 3
+assert new_meta in instances
+assert new_inst1 in instances
+assert new_inst2 in instances
+
+uuids = s.get_uuids()
```
CasperWA, line 59:
```suggestion
uuids = storage.get_uuids()
```

CasperWA (commented):
Looks good to me.

Not that it's relevant now, but I just found that someone is developing mongomock also for the asynchronous client ([mongomock-motor](https://github.com/michaelkryukov/mongomock_motor)).

storages/python/python-storage-plugins/mongodb.py
jesper-friis: Only reason id to make the code easier to read by reducing line lengths.
Moved `self.options = opts` to the end of the function.

storages/python/python-storage-plugins/mongodb.py
```diff
@@ -0,0 +1,92 @@
...
+            @mongomock.patch(servers=((mongo_url.hostname, port), ))
+            def get_client():
+                return open_client()
+        else:
+            def get_client():
+                return open_client()
+
+        def open_client():
+            client = pymongo.MongoClient(
+                host=uri,
```
CasperWA, line 60: Do you have an example of what `uri` might look like?
In PyMongo terms, the Mongo URI is the full URI, e.g., `mongodb://localhost:27017/`, where `host` would then be `localhost` in this case.

Here, you are using the `uri` parameter as the `host` (+ `port`), which clashes with the PyMongo definition.

This is most likely fine, but it would be good with a `uri` parameter example in the method's doc string or a reference to a more expanded documentation of how this parameter may look like?
This whole thing will e.g., _not_ work if I include the scheme part of the URI when using the mongodb storage backend...

jesper-friis: Good point. `location` would probably be a better name for the `uri` argument. It might be worth adding an issue for changing it consistently throughout all plugins?

If the user wants to fetch an instance from a MongoDB storage, he or she could write

```python
inst = Instance.from_url('mongodb://localhost:27017?user=me#4f3f2898-69e3-5a0d-820a-c4e87987dbce')
```

which would result in the following three calls to the mongodb plugin: 

```python
mongodb.open(uri='localhost:27017', options='user=me')
mongodb.load(id='4f3f2898-69e3-5a0d-820a-c4e87987dbce')
mongodb.close()
```

CasperWA: An issue should be raised for this to improve the legibility of the MongoDB plugin code.

CasperWA (approved):
Looks good, please create an issue according to the comment I raised above.

storages/python/python-storage-plugins/mongodb.py
jesper-friis: Not sure what happened with my previous reply. But I have now added issue #325 as well as an additional point to issue #80. Merging now...

PR #329, opened 2022-08-27
Added OntoPy triplestore backend

jesper-friis:


bindings/python/tests/check_import.py
```diff
@@ -0,0 +1,30 @@
...
+
+    Returns:
+        A module object if the module can be imported.  Otherwise None is
+        returned.
+    """
+    try:
+        module = importlib.import_module(module_name, package)
+    except ImportError:
+        if skip:
+            sys.exit(44)
```
daniel-sintef, line 25: Why 44 error? Maybe this should be mentioned in the doc string. "Tries to import, else returns a 44 <description> error

bindings/python/tests/test_triplestore.py
```diff
@@ -103,89 +97,113 @@
...
+        EX.MyConcept, "http://onto-ns.com/meta/0.1/MyEntity", "myprop")
+    ts.add((EX.MyConcept, RDFS.subClassOf, OWL.Thing))
+    ts.add((EX.AnotherConcept, RDFS.subClassOf, OWL.Thing))
+    ts.add((EX.Sum, RDFS.subClassOf, OWL.Thing))
+    assert ts.has(EX.Sum) == True
+    assert ts.has(EX.Sum, RDFS.subClassOf, OWL.Thing) == True
+    assert ts.has(object=EX.NotInOntology) == False
+
 
+    def sum(a, b):
```
daniel-sintef, line 116: "Summarise" seems a bit strange, is this the best term?

bindings/python/triplestore/backends/ontopy.py
```diff
@@ -0,0 +1,197 @@
...
+                    os.remove(filename)
+
+    def query(self, query_object, native=True, **kwargs):
+        """SPARQL query."""
+        if native:
+            res = self.onto.world.sparql(query_object)
+        else:
+            graph = self.onto.world.as_rdflib_graph()
+            res = graph.query(query_object, **kwargs)
+        # TODO: Convert result to expected type
```
daniel-sintef, line 188: Is this TODO important?

daniel-sintef: The TODO should be removed and placed into a github issue

bindings/python/triplestore/backends/rdflib.py
```diff
@@ -93,6 +93,7 @@ def serialize(self, destination=None, format='turtle', **kwargs):
 
     def query(self, query_object, **kwargs):
         """SPARQL query."""
+        # TODO: convert to returned object
```
daniel-sintef, line 96: TODO should be removed and placed as github issue

bindings/python/tests/check_import.py
jesper-friis: You are right, we should have chosen error 42 :-)

Joke aside, it is a specific error code that tells cmake to skip the test. A comment is now added

bindings/python/tests/test_triplestore.py
jesper-friis: Probably not. Changed to "Returns the sum of `a` and `b`."

bindings/python/triplestore/backends/ontopy.py
jesper-friis: Agree that it should be made an issue, but I don't think that we should remove the TODO - it makes it easier to spot where the needed changes should go

jesper-friis: Added issue #336

bindings/python/triplestore/backends/ontopy.py
```diff
@@ -0,0 +1,197 @@
...
+
+    from dlite.triplestore import Triple
+
+
+class OntopyStrategy:
+    """Triplestore strategy for EMMOntoPy.
+
+    Arguments:
+        onto: Ontology to initiate with.  Defaults to an new ontology with
+            `base_iri`.
```
francescalb, line 24:
```suggestion
        base_iri: The base iri of the ontology.
        onto: Ontology to initiate with.  Defaults to an new ontology with
            `base_iri`.
```

bindings/python/triplestore/backends/ontopy.py
```diff
@@ -0,0 +1,197 @@
...
+        load: Whether to load the ontology.
+        kwargs: Keyword arguments passed to the ontology load() method.
+    """
+    def __init__(
+            self,
+            base_iri: str,
+            onto: Ontology = None,
+            load: bool = False,
+            kwargs: dict = {}
```
francescalb, line 33: Here we require that base_iri is given. Should we not be able to get the base_iri from the ontology if we load an external one.

bindings/python/triplestore/backends/ontopy.py
```diff
@@ -0,0 +1,197 @@
...
+                self.onto._del_data_triple_spod(s, p, o, d)
+            else:
+                self.onto._del_obj_triple_spo(s, p, o)
+
+    # Optional methods
+    def parse(self, source=None, location=None, data=None, format=None,
+              encoding=None, **kwargs):
+        """Parse source and add the resulting triples to triplestore.
+
+        The source is specified using one of `source`, `location` or `data`.
```
francescalb, line 116: In the current implementation it will choose the one it finds first in the order source, location, data. This means that if more than one is given it will only use the first it finds. IF that is desired the order of preference should be added in the docstring.

Alternatively, if only one can be given, definig more than one should throw and error.

bindings/python/tests/test_triplestore.py
```diff
@@ -103,89 +97,113 @@
...
+        "ontopy", base_iri="http://onto-ns.com/ontologies/examples/food",
+    )
+    ts4.parse(ontopath_food)
+
+    ts5 = Triplestore(
+        "ontopy", base_iri="http://onto-ns.com/ontologies/examples/food",
+    )
+    ts5.bind('food', FOOD)
+    with open(ontopath_food, 'rt') as f:
+        ts5.parse(data=f.read())
```
francescalb, line 209: Several functions that belong to the ontopy triplestore backend are not tested: 
add_triples, remove, query, triples, remove, serialize and update. 

These need to be added.

francescalb (changes requested):
Nice cakend. Please add the the tests that are missing for the ontopy backend. Also some examples of usage in the documentation would be very useful for a potential user (like me).

bindings/python/tests/test_triplestore.py
jesper-friis: We will add more tests in https://github.com/SINTEF/dlite/issues/341

bindings/python/triplestore/backends/ontopy.py
jesper-friis: The documentation says "_The source is specified using one of `source`, `location` or `data`._", so I think it is fair to consider providing more than one is undefined behaviour. That leaves little more freedom for the implementation of the backends. OK?

bindings/python/triplestore/backends/ontopy.py
jesper-friis: Thanks for the correction. I also improved the documentation of `onto`

bindings/python/triplestore/backends/ontopy.py
jesper-friis: Fixed

bindings/python/triplestore/backends/ontopy.py
francescalb: OK. But I still think and error or warning should be raised if more than one of them is given.

bindings/python/tests/test_triplestore.py
francescalb: OK

bindings/python/triplestore/backends/ontopy.py
jesper-friis: A check has been pushed

jesper-friis:
> Nice backend. Please add the the tests that are missing for the ontopy backend. Also some examples of usage in the documentation would be very useful for a potential user (like me).

Ok to leave the implementation of the missing tests to issue #341, which will require a total reorganisation of the existing triplestore tests?

francescalb (approved):
Looks good.

PR #340, opened 2022-09-19
Qudt prefix

jesper-friis:
Added two additional pre-defined prefixes to the triplestore module:
```
qudtu: <http://qudt.org/vocab/unit/>
om: <http://www.ontology-of-units-of-measure.org/resource/om-2/>
```

Question: Should we use "qudt" as prefix instead of "qudtu", even though that there exists other QUDT ontologies?


bindings/python/triplestore/triplestore.py
```diff
@@ -203,6 +203,8 @@ def __add__(self, other):
 DCAT = Namespace("http://www.w3.org/ns/dcat#")
 TIME = Namespace("http://www.w3.org/2006/time#")
 FNO = Namespace("https://w3id.org/function/ontology#")
+QUDTU = Namespace("http://qudt.org/vocab/unit/")
```
quaat, line 206: This is fine, however QUDT uses the prefix _unit_:
 @prefix unit: <http://qudt.org/vocab/unit/> .

quaat (approved):
Minor comment

bindings/python/triplestore/triplestore.py
jesper-friis: Maybe it is better that we also uses `UNIT` as prefix?

PR #342, opened 2022-09-20
WIP: Pint unit registry from ontologies

### LINKED ISSUE #320: Populate unit DB from ontologies
dlite.mappings currently uses pint with its built-in unit database. We want to replace the pint built-in unit database with ontologies.

Create a new Python module that exposes a `ureg.Quantity` class, where `ureg` is a pint UnitRegistry populated from ontologies. 

For efficiency the ontologies could be cached to a local pint ureg file.

Some unit ontologies that would be nice to support include EMMO, QUDT, OM and wikidata (has fewer errors than QUDT according to Simon Stier).

ajeklund:
Population of the Pint unit registry from unit ontologies, including QUDT and EMMO.

Closes #320.


bindings/python/triplestore/units.py
```diff
@@ -0,0 +1,250 @@
...
+    return ts
+
+def parse_qudt_dimension_vector(dimension_vector: str) -> dict:
+    dimensions = re.findall(r'[AELIMHTD]-?[0-9]+', dimension_vector)
+    
+    result = {}
+    for dimension in dimensions:
+        result[dimension[0]] = dimension[1:]
+
+    expected_keys = ["A", "E", "L", "I", "M", "H", "T", "D"]
```
jesper-friis, line 27:
```suggestion
    expected_keys = "AELIMHTD"
```

bindings/python/triplestore/units.py
```diff
@@ -0,0 +1,250 @@
...
+    for letter in expected_keys:
```
jesper-friis, line 28:
```suggestion
    for letter in "AELIMHTD":
```

bindings/python/triplestore/units.py
```diff
@@ -0,0 +1,250 @@
...
+        if letter not in result.keys():
+            raise Exception("Missing dimension \"" + letter + "\" in dimension vector " + dimension_vector)
```
jesper-friis, line 30: It is a good practice to keep lines below 79 characters

bindings/python/triplestore/units.py
```diff
@@ -0,0 +1,250 @@
...
+
+"""
+from email.policy import default
+from pint import UnitRegistry, Quantity
+import re
+from triplestore import Triplestore, RDFS
+import warnings
+
+def load_qudt():
+    print("Loading QUDT unit ontology.")
```
jesper-friis, line 13: Avoid print statements in libraries. Use the logging module when needed.

In this case it is fine to just remove the print statement

bindings/python/triplestore/units.py
```diff
@@ -0,0 +1,250 @@
...
+    ts = Triplestore(name="rdflib")
+    ts.parse(source="http://qudt.org/2.1/vocab/unit")
+    ts.parse(source="http://qudt.org/2.1/schema/qudt")
+    print("Finished.")
```
jesper-friis, line 17:
```suggestion
```

bindings/python/triplestore/units.py
```diff
@@ -0,0 +1,250 @@
...
+    kilogram
+    ampere
+    kelvin
+    mole
+    candela
+@end""")
+    return result
+
+
+# Test code.
```
jesper-friis, line 87: Move test code into a separate test script.

bindings/python/triplestore/units.py
```diff
@@ -0,0 +1,250 @@
...
+        exponent = dimension_dict[letter]
+        if int(dimension_dict[letter]) < 0:
+            result += "/ " + base_units[letter] + "**" + exponent[1:] + " "
+        elif int(dimension_dict[letter]) > 0:
+            result += "* " + base_units[letter] + "**" + exponent + " "
+    return result
+
+def pint_SI_base_units_definition() -> list:
+    # SI units as defined in the pint default registry: pint/default_en.txt
+    result = []
```
jesper-friis, line 61: Faster and easier code if you just write
```python
    result = [
        "meter = [length] = m = metre",
        "second = [time] = s = sec",
         ...
    ]
```

bindings/python/triplestore/units.py
```diff
@@ -0,0 +1,250 @@
...
+        "L": "m",
+        "I": "cd",
+        "M": "kg",
+        "H": "K",
+        "T": "s",
+        "D": "1",
+    }
+
+    result = ""
+    for letter in base_units:
```
jesper-friis, line 51: Easier to just write
```
for letter, unit in base_units.items():
    ...
```

.gitignore
```diff
@@ -32,3 +32,4 @@ out
 
 # Generated build docker files
 Dockerfile-*linux*[^.template]
+test_output.txt
```
jesper-friis, line 35: Don't pollute the root .gitignore file unnecessarily...
```suggestion
```

jesper-friis:
The file `bindings/python/triplestore/__pycache__/triplestore.cpython-310.pyc` should not be checked in.

bindings/python/triplestore/units.py
ajeklund: Deleted this method since I do not use it (implemented code which handles the SI base units as loaded from QUDT).

bindings/python/triplestore/units.py
jesper-friis:
```suggestion
```

bindings/python/triplestore/units.py
ajeklund: Thank you, now implemented as suggested.

jesper-friis:
Binary .pyc files should not be added to the repo.

jesper-friis:
Some general comments: 
- pint understands prefixes, so we don't need to register 'Megapascal' if 'pascal' is already registered
- prefixes should be added to the unit registry
- by default pint writes units in lower case, I think we should do the same (but could possible add the upper-case variant as an alias)
- when available, use the standard unit name instead of the QUDT name as canonical name. For instance, Pascal should be registered as follow:

      pascal = 1.0 / m**1 * kg**1 / s**2  = Pa = PA = http://qudt.org/vocab/unit/PA

  or maybe

      pascal = 1.0 / m**1 * kg**1 / s**2  = Pa = Pascal = PA = http://qudt.org/vocab/unit/PA

- I think conflicts in the QUDT names ('PA' and possible 'Pascal' in the example above) should be resolved by skipping these names in the registry such that Pascal would be registed as

      pascal = 1.0 / m**1 * kg**1 / s**2  = Pa = http://qudt.org/vocab/unit/PA

- use `logging.warning()` instead of `warnings.warn()` in cases where the user can't do anything about the situation, but the event should still be noted. `warnings.warn()` is for the cases where the user should take action to eliminate the warning.

jesper-friis:
We don't want to include the `bindings/python/triplestore/__pycache__/` directory in the repo.

jesper-friis (approved):
Approved if you remove the `test_output.txt` file and `__pycache__` directory from the git repo.

In this case it is important to squash merge to avoid that these huge files leaks into the history making the repo unnecessarily large.

jesper-friis (approved):
Approved if you remove the test_output.txt file and __pycache__ directory from the git repo and all tests runs successfully.

In this case it is important to squash merge to avoid that these huge files leaks into the history making the repo unnecessarily large.

PR #345, opened 2022-09-20
Updated README by removing optional when dependency is needed fo build.

francescalb:


README.md
```diff
@@ -290,17 +290,17 @@ The sources can be cloned from GitHub
...
-  - NumPy development libraries, optional (needed by Python bindings)
-  - [SWIG v3][10], optional (needed by building Python bindings)
-  - [Doxygen][11], optional, used for documentation generation
-  - [valgrind][12], optional, used for memory checking (Linux only)
-  - [cppcheck][13], optional, used for static code analysis
-  - [mongomock][mongomock], optional, used for testing mongodb storage plugin
-
+  - [cmake][9], required for building - note that cmake isntalled from pypi does not always work.
+  - hdf5 development libraries, needed by HDF5 storage plugin.
+  - Python 3 development libraries, needed by Python bindings.
```
kriwiik, line 295: Add link or exact name to be used for apt-get install command?

README.md
```diff
@@ -290,17 +290,17 @@ The sources can be cloned from GitHub
...
+  - NumPy development libraries, needed by Python bindings.
+  - [SWIG][10] needed by building Python bindings.
```
kriwiik, line 297: Needed for building, or needed by python bindings?

kriwiik (changes requested):
* The git clone command provided by the readme includes a code.sintef.no address. This does not work (at least not for me). Change to the correct address. 

`Setting up the environment` section
* Fix typo: As a developeR (now says developed)
* Add explanations for the "command blocks" where the environment is activated and the Python3_ROOT etc variables are set - what are we doing here and why? 
* In the "command block" under "build dlite", remove one of the "-DPython3_LIBRARY=..." 

Also, see comments.

README.md
francescalb: Needed for building with python bindings I think.

README.md
francescalb: No, this is dependent on the linux distribution. I think we can assume that the person installing dlite for building person knows this for the given system or is able to google it.

README.md
```diff
@@ -418,16 +417,24 @@ variables](doc/environment_variables.md) for more details.
 An example of how to install dlite as developer within a python environment 
 in linux is given below.  Make sure that all required dependencies
 are installed within the environment.
+
+First activate the environment, e.g.:
 ```console	
 source /path/to/dedicated/pythonenvironment/bin/activate
 ```
+Set the Python variables. The followig should automatically
```
kriwiik, line 425:
```suggestion
Set the Python variables. The following should automatically 
```

kriwiik (approved):
Approved after typo is fixed (see suggestion for line 425)

PR #348, opened 2022-09-22
Improved error codes

daniel-sintef:
Please take a look at the dlite-errors.h and the new FAILCODE macros


jesper-friis (approved):
This is a nice and clean PR done in the right way. 
A good start that can be merged as is.
I like the introduction of FAILCODE() allowing incremental implementation of the PR.

PR #349, opened 2022-09-23
Added clear description of expected return value from the Triplestore.query() method

jesper-friis:


bindings/python/triplestore/backends/rdflib.py
```diff
@@ -92,8 +92,17 @@ def serialize(self, destination=None, format='turtle', **kwargs):
...
+
+        Parameters:
+            query_object: String with the SPARQL query.
+            kwargs: Keyword arguments passed to rdflib.Graph.query().
+
+        Returns:
+            List of tuples of IRIs for each matching row.
+        """
+        rows = self.graph.query(query_object=query_object, **kwargs)
+        return [tuple(str(v) for v in row) for row in rows]
```
quaat, line 105: Should the query API be restricted to SELECT statements? If not we might have to consider returning a Results-object?

bindings/python/triplestore/interface.py
```diff
@@ -48,7 +48,15 @@ def serialize(self, destination=None, format='xml', **kwargs)
         """
 
     def query(self, query_object, **kwargs)
```
quaat, line 50: In the interface, the return-type should ideally be defined as well

quaat (commented):
Please consider the questions asked in the review. If there is a need to follow up, please create separate (new) issues for these. This branch may be merged as is.

bindings/python/triplestore/backends/rdflib.py
jesper-friis: Good point. INSERT and DELETE queries are expected to be used with the update() method since they modifies the triplestore. We should document that. As I understand is this how rdflib does it.

What other types of queries does there exists?

bindings/python/triplestore/interface.py
jesper-friis: It is documented on line 51-59

bindings/python/triplestore/backends/rdflib.py
jesper-friis: I have updated the docstrings

bindings/python/triplestore/interface.py
quaat: I was not talking about documentation, but rather type hints

quaat (approved):

quaat (commented):
looks good

PR #355, opened 2022-10-15
Fixed gcc warnings

### LINKED ISSUE #354: Fix new warnings from gcc 12.2.1
gcc 12.2.1 issues two new warnings:

A `format-truncation` warnings in `mu_assert_string_eq()`. To silence it, we have to check the return value of `snprintf()` even though it is not used. Not sure that it improves the code... but it is required for successfully compiling dlite with gcc 12.2.1 since all warnings are turned into errors.

The second fix is a real improvement for handling the return value of `realloc()` correctly.

jesper-friis:
Closes #354


daniel-sintef (approved):
Can you clarify what you did in this commit and why it worked?

PR #365, opened 2022-11-23
Build against python3.7

jesper-friis:
Added cmake configuration file for building dlite against python3.7 in a virtual environment.

Good to have for a later time since it is not trivial to do...


build-python3.7.cmake
```diff
@@ -0,0 +1,42 @@
...
+# -----------------------------
+# With virtualenvwrapper installed, do
+#
+#     mkvirtualenv dlite37 -p /usr/bin/python3.7
+#     python3.7 -m pip install -U pip
+#     python3.7 -m pip install -r requirements.txt
+#     python3.7 -m pip install -r requirements_dev.txt
+#
+# 3. Build dlite
+# --------------
```
CasperWA, line 21: Should there be a step to also activate the new virtualenv, or is this all taken care of in the commands below to use the virtualenv interpreter?

CasperWA (approved):
Looks good to me, however, please see the comment.

build-python3.7.cmake
jesper-friis: Good point. But `mkvirtualenv` automatically activates the new environment, so it is already taken care of.

PR #367, opened 2022-11-24
366 generate python api documentation

quaat:
Generates python-documentation with Sphinx from CMake

PR #370, opened 2022-11-25
Update CD - Docs workflow

CasperWA:
Use the new Sphinx-powered documentation introduced in #367.

This will generate a GH Pages artifact, which can be used as the documentation to be published instead of a dedicated branch.


CasperWA:
@quaat Please check the path is correct as specified in line 79 of the `cd_docs.yml` file.

jesper-friis (approved):
Approved given that the checks succeeds (still running while doing this review...)

PR #374, opened 2022-12-11
Added `metaid` argument to Collection.get_instances()

jesper-friis:
Makes it easy to get all instances of a given metadata in a collection.


bindings/python/dlite-collection-python.i
```diff
@@ -7,23 +7,23 @@
...
+        if self.rettype == 'T':  # return (s,p,o) tuple
             return (v.s, v.p, v.o)
-        elif self.rettype == 's':
+        elif self.rettype == 's':  # return subject
             return v.s
-        elif self.rettype == 'p':
+        elif self.rettype == 'p':  # return predicate
             return v.p
-        elif self.rettype == 'o':
+        elif self.rettype == 'o':  # return pbject
```
sygout, line 26: return object

bindings/python/dlite-collection-python.i
```diff
@@ -194,9 +194,20 @@ class Collection(Instance):
...
-        """Returns a generator over all instances in collection."""
-        return _CollectionIter(self, s=None, p=None, o=None, rettype='I')
+    def get_instances(self, metaid=None):
+        """Returns a generator over all instances in collection.
+        If `metaid` is given, only instances of this metadata will be
+        returned.
+        """
+        iter = _CollectionIter(self, s=None, p=None, o=None, rettype='I')
+        if metaid:
+            uri = metaid.uri if hasattr(metaid, "uri") else metaid
```
sygout, line 204: add comment about what you are doing and the example of the two cases

bindings/python/tests/test_collection.py
```diff
@@ -114,6 +114,12 @@
 i1, = coll.get_instances()
 assert i1 == inst1
 
+# We have no collections in the collection
```
sygout, line 117: why do we do this test? why would it be a problem?

jesper-friis: It is not a problem, we just test that get_instances() behave as expected.

On line 118 we test get_instances() with a metaid referring to a collection. Since we have no nested collection in the collection, the returned iterator should be empty.

bindings/python/tests/test_collection.py
```diff
@@ -114,6 +114,12 @@
...
+assert not list(coll.get_instances(dlite.COLLECTION_ENTITY))
+
+i1, = coll.get_instances('http://onto-ns.com/meta/0.1/MyEntity')
```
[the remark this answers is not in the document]
jesper-friis: That is tested on lines 114-115.

PR #378, opened 2022-12-12
Give instances a default uri

### LINKED ISSUE #377: Give instances a default uri
When an instance haven't got a default uri, it will default to its metadata uri followed by "/" followed by its uuid, like

    http://onto-ns.com/meta/0.1/Energy/5580a9e5-6692-41f5-8f5d-3083aec76b8d

This require that we update getuuid() such that it recognises id's on the above form and returns the uuid part instead of generating an uuid version 5 hash from it.

jesper-friis:
Closes #377


doc/concepts.md
```diff
@@ -1,36 +1,55 @@
...
+The tool `dlite-getuuid` can be used to manually convert URIs to their
 corresponding UUIDs.
 
-If the `id` provided to dlite_datamodel() or dlite_instance_create()
-is not `NULL` or a valid UUID, it is interpreted as an unique uri
-referring to the instance.  dlite will then generate a version 5
-UUID from the `id` (from the SHA-1 hash of `id` and the DNS namespace)
-that the instance will be stored under.  The original `id` will also be
-stored and can be retrieved with dlite_datamodel_get_meta_uri().
+A resent feature in DLite is that data instances with no user-defined
```
lovfall, line 38: maybe not use resent, as the documentation in principle should last for a long time

src/getuuid.c
```diff
@@ -1,27 +1,71 @@
...
+  if (*(s++) != '-') return 0;
+  for (i=0; i<12; i++) if (!isxdigit(*(s++))) return 0;
+  return 1;
+}
+
+/* Returns non-zero if `s` matches <URI>/<UUID>. `len` is the length of `s`.
+   An optional final hash or slash will be ignored.*/
+int isinstanceuri(const char *s, int len)
+{
+  if (!len) len = strlen(s);
```
lovfall, line 36: If I read this correctly, you sort of make the length of s optional? Does that make sense? Would it not be better to either raise an error, or remove the argument and always calculate the length?

src/getuuid.c
```diff
@@ -1,27 +1,71 @@
...
- * is generated from `id` using the DNS namespace.
- * Otherwise `id` is copied to `buff`.
- *
- * Length of `buff` must at least 37 bytes (36 for UUID + NUL termination).
- *
- * Returns the UUID version, if a new UUID is generated or zero if
- * `id` is simply copied.  On error, -1 is returned.
+  Writes an UUID to `buff` based on `id`.
+
+  It follow the follow heuristics:
```
lovfall, line 51: If follows the following...?

src/getuuid.c
```diff
@@ -1,27 +1,71 @@
...
+  - If `id` is NULL or empty, a new random version 4 UUID is generated.
+    Return: UUID_RANDOM
+  - If `id` is a valid UUID, it is copied as-is to `buff`.
+    Return: UUID_COPY
+  - If `id` matches `<URI>/<UUID>` then it returns the <UUID> part.
+    `id` may optionally end with a final hash (#) or slash (/), which will
+    be ignored.
+    Return: UUID_EXTRACT
+  - Otherwise is `id` is an invalid UUID string.  A new version 5 sha1-based
```
lovfall, line 60: Otherwise if ... ?

src/getuuid.c
```diff
@@ -41,15 +85,24 @@ int getuuidn(char *buff, const char *id, size_t len)
...
+              (len == UUID_LEN+1 && strchr("#/", id[len]))) &&
+             isuuid(id)) {
     strncpy(buff, id, UUID_LEN);
     buff[UUID_LEN] = '\0';
-    version = 0;
+    version = UUID_COPY;
+  } else if (isinstanceuri(id, len)) {
+    if (id[len-1] && strchr("#/", id[len-1])) len--;
+    len -= UUID_LEN;
+    assert(len > 0);
```
lovfall, line 99: Why assert here?

src/getuuid.h
```diff
@@ -9,18 +9,35 @@
...
+  It follow the follow heuristics:
+  - If `id` is NULL or empty, a new random version 4 UUID is generated.
+    Return: UUID_RANDOM
+  - If `id` is a valid UUID, it is copied as-is to `buff`.
+    Return: UUID_COPY
+  - If `id` matches `<URI>/<UUID>` then it returns the <UUID> part.
+    `id` may optionally end with a final hash (#) or slash (/), which will
+    be ignored.
+    Return: UUID_EXTRACT
+  - Otherwise is `id` is an invalid UUID string.  A new version 5 sha1-based
```
lovfall, line 32: first is = if?

src/utils/tests/test_strutils.c
```diff
@@ -187,6 +187,80 @@ MU_TEST(test_strhex_decode)
...
+  mu_assert_int_eq(strcatSpecific, strcategory('+'));
+  mu_assert_int_eq(strcatSpecific, strcategory(','));
+  mu_assert_int_eq(strcatSpecific, strcategory(';'));
+  mu_assert_int_eq(strcatSpecific, strcategory('='));
+  mu_assert_int_eq(strcatPercent, strcategory('%'));
+  mu_assert_int_eq(strcatNul, strcategory('\0'));
+  mu_assert_int_eq(strcatOther, strcategory('"'));
+  mu_assert_int_eq(strcatOther, strcategory('`'));
+  mu_assert_int_eq(strcatOther, strcategory('<'));
+  mu_assert_int_eq(strcatOther, strcategory('>'));
```
lovfall, line 225: What happens if the character is non-7-bit ascii?

lovfall (changes requested):
See comments

doc/concepts.md
jesper-friis: Documentation updated

src/getuuid.c
jesper-friis: You read it correctly. You are right, it doesn't make much sense (except for simplifying the tests) since this is an internal help function which probably should have been declared static. The only reason that it is not static is because that would make it inaccessible for unit testing...

src/getuuid.c
jesper-friis: thanks

src/getuuid.c
jesper-friis: It is just a safety guard. If the code works as intended, `len` should always be tositive. 

If `id` is an instance URI of the form `<meta.uri>/<uuid>` then `len` should be positive (larger than 9, I think). What I really don't want is that `len` for some unknown reason becomes negative, resulting in accessing some random memory. Then I prefer aborting dlite loudly...

src/getuuid.h
jesper-friis: second "is" should be removed

src/utils/tests/test_strutils.c
jesper-friis: It should be an other. Added a test to confirm it :-)

PR #380, opened 2022-12-14
Alternative to hack #379

jesper-friis:
This PR builds on top of #378 

This solution simply provide a method Instance.get_uri() that always return an URI.


src/getuuid.c
```diff
@@ -1,27 +1,71 @@
...
 
+#include "utils/strutils.h"
 #include "utils/uuid.h"
 #include "utils/uuid4.h"
 #include "getuuid.h"
 
 
+/* Returns non-zero if `s` is a valid UUID. */
+
+int isuuid(const char *s)
```
quaat, line 17: I asked ChatGPT to come up with code to validate an UUID. Here is what it suggested:

```
#include <stdbool.h>
#include <stdio.h>
#include <string.h>
#include <stdlib.h>

bool is_valid_uuid(const char *uuid) {
  // UUID strings are always 36 characters long
  if (strlen(uuid) != 36) {
    return false;
  }

  // UUID strings consist of 32 hexadecimal characters, with four hyphens
  // inserted at specific positions to separate the different parts of the
  // UUID.
  for (int i = 0; i < 36; i++) {
    if (i == 8 || i == 13 || i == 18 || i == 23) {
      // The character at these positions should be a hyphen
      if (uuid[i] != '-') {
        return false;
      }
    } else {
      // The character at all other positions should be a hexadecimal digit
      if (!isxdigit(uuid[i])) {
        return false;
      }
    }
  }

  // If we reach this point, the UUID is valid
  return true;
}
```

quaat (approved):

src/getuuid.c
jesper-friis: Nice code :-)

PR #381, opened 2022-12-18
Do not build documentation by default

jesper-friis:
The standard user would normally use the online documentation rather than building it from scratch.

It therefore make sense to change the default to not building documentation. This also removes some development dependencies and make the build process less error prone.


CMakeLists.txt
```diff
@@ -61,7 +61,7 @@ option(WITH_FORTRAN     "Whether to build Fortran bindings"                 NO)
 option(WITH_HDF5        "Whether to build with HDF5 support"               YES)
 option(WITH_REDLAND     "Whether to build with RDF support (using librdf)" YES)
 option(WITH_THREADS     "Whether to build with threading support"           NO)
-option(WITH_DOC         "Whether to build documentation using doxygen"     YES)
+option(WITH_DOC         "Whether to build documentation using doxygen"      NO)
```
CasperWA, line 64: Is this still only explicitly to build with doxygen, or does it now include the Sphinx build? If that latter is true, I would update the comment here, changing `doxygen` to `Sphinx`.

jesper-friis: This include both the doxygen and Sphinx documentation. Changed it to sphinx+doxygen

CasperWA (approved):

CasperWA (changes requested):
Update the README where it discusses building the documentation.

CasperWA (changes requested):

README.md
```diff
@@ -381,6 +381,11 @@ system defined in Dockerfile.
 
 ### Build documentation
 
+In order to reduce build dependencies for the causal user, DLite does not
+build documentation by default.  Provide the `-DWITH_DOC=YES` option to
+`cmake` to build the documentation.
```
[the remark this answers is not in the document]
jesper-friis: The `-DWITH_DOC=YES` option should be provided when you do the initial configuration of cmake. The sub-sections below are for the build stage. However, on line 337 we already show an example of providing the `-DWITH_DOC=YES` at the cmake configuration stage.

CasperWA (approved):

Issue #352, opened 2022-10-13
No identification of error in datamodel


francescalb:
If there is an error in the datamodel, the error reported is that is cannot find it. 
This is very confusing, because it does acutally find it but it cannot instantiate it. 
There should be an error saying "Cannot find it" and one that is "Can find the uri, but cannot instantiate it"
or similar.

francescalb:
An example. If you do not have need for "dimensions" (all dims are 1), then it is still needed to have an empty list for "dimensions". If it is missing, the error returned is :   

File "/usr/local/lib/python3.9/site-packages/dlite/dlite.py", line 1512, in get_instance
    raise DLiteError(f"no such instance: {id}")
    
   The error is actually that 'dimensions" was not added in the datamodel itself.

PR #385, opened 2023-01-01
Improve error reporting for missing dimensions, properties and/or relations

### LINKED ISSUE #353: Datamodel dimensions
When adding dimensions to the datamodel that are not used in the properties, no error is returned. 
This should not be valid. I suggest checking that all dimensions added are actually made use for, if not the datamodel should not contain it.

jesper-friis:
Closes #353 

This solution may not be perfect, but hopefulle it is an improvement. A missing dimension should now result in the following message:

    DLiteError: Error 1: metadata does not confirm to schema, please check dimensions, properties and/or 
    relations: http://onto-ns.com/test/0.1/ch

It is understandable? It is possible to add more finegrained checks for missing/unexpected dimensions, properties and/or relations, but it will require more code.


jesper-friis:
Referred to wrong issue - should have been #352. Closing this PR and add a new one with updated branch name.

PR #404, opened 2023-01-03
Added C-API documentation to the sphinx documentation project. TODO: …

quaat:
Added C-API documentation to the sphinx documentation project. TODO: migrate pydoc and doc (need a follow-up task)
Other possible improvements is a script that parse the generated index.xml-file and generate the .rst files and folders necessary


jesper-friis (approved):
Looks nice! What remains should be followed up in new taks. This include (among others):
- make a user manual based on the markdown files in the doc/ directory
- make a Python reference manual
- make a C reference manual with the content of this PR
- go through and possibly fix the errors and warnings from the documentation building step
- figure out why some documentation does not get included, and include it. This is e.g. the case for:
  - the description at the top of src/dlite-entity.h
  - the src/utils/README.md file
  - probably more...

PR #414, opened 2023-01-04
Made install rule for FORTRAN modules when FORTRAN bindings are built

lovfall:
This fixes the problem with missing modules when FORTRAN modules are built and installed

### DIFF
--- a/bindings/fortran/CMakeLists.txt
+++ b/bindings/fortran/CMakeLists.txt
@@ -16,6 +16,20 @@ target_link_libraries(dlite-fortran
   )
 
 
+## create list of modules to install from ${sources}
+set(modules "")
+foreach(source ${sources})
+  string(REPLACE ".f90" ".mod" module ${source})
+  if(EXISTS ${CMAKE_CURRENT_BINARY_DIR}/${module})
+      list(APPEND modules ${CMAKE_CURRENT_BINARY_DIR}/${module})
+  endif()
+endforeach()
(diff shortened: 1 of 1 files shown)


bindings/fortran/CMakeLists.txt
```diff
@@ -16,6 +16,20 @@ target_link_libraries(dlite-fortran
   )
 
 
+## create list of modules to install from ${sources}
+set(modules "")
+foreach(source ${sources})
+  string(REPLACE ".f90" ".mod" module ${source})
+  if(EXISTS ${CMAKE_CURRENT_BINARY_DIR}/${module})
```
jesper-friis, line 23: This foreach look will run at the configuration stage. At that stage we would have no modules in the binary dir, and will therefore not add them to `modules`.

Can't we simply remove the if-statement?

lovfall: The foreach loop works, but I think we can remove the if-statement yes, it should be redundant. There was a bug in the first commit, will make an update

jesper-friis: Yes, the if-statement must be removed, otherwise the modules will not be installed (unless you reconfigure and install after you have first build the modules once).

jesper-friis (changes requested):
The if-statement must be removed, otherwise the modules will not be installed (unless you reconfigure and install after you have first build the modules once).

lovfall:
@jesper-friis: There are no if-statements to remove, so the change request is void. Did you look at the latest commit?

jesper-friis (approved):

PR #422, opened 2023-01-04
Do not install examples into the <prefix>/share/dlite/ directory

### LINKED ISSUE #412: Remove Py files under built `mapping-plugins` folder
These come from examples and should not be part of the "installed" plugins.

jesper-friis:
Closes #412


CasperWA (approved):
Perfect!

PR #424, opened 2023-01-04
[DO NOT MERGE] Reverting the changes that breaks the postgresql tests

jesper-friis:
The changes in test_postgresql_storage_python.py makes 

         69 - test_postgresql_storage (Failed)
         70 - test_postgresql_storage2 (Failed)
         72 - test_postgresql_storage_python (Failed)

fail. The two first with 

    Error 1: error calling postgresql.load(): RuntimeError: Fetching ('http://onto-ns.com/meta/0.1/Person',) from PostgreSQL database results in a different meta ID to be returned: http://onto-ns.com/meta/0.1/Person

and the last with

    Traceback (most recent call last):
      File "/home/jesperf/prosjekter/software/dlite/storages/python/tests-python/test_postgresql_storage_python.py", line 45, in <module>
        open_start = lines.index(df + 'open(self, uri, options=None):\n')
    ValueError: '    def open(self, uri, options=None):\n' is not in list

The changes in test_postgresql_storage_python.py makes test_postgresql_storage_python fail with 

    Running Python test <test_postgresql_storage_python.py>...
    Traceback (most recent call last):
    File "/home/jesperf/prosjekter/software/dlite/storages/python/tests-python/test_postgresql_storage_python.py", line 42, in <module>
        open_start = lines.index(df + 'open(self, uri: str, options: "Optional[str]" = None) -> None:\n')
    ValueError: '    def open(self, uri: str, options: "Optional[str]" = None) -> None:\n' is not in list

### DIFF
--- a/storages/python/python-storage-plugins/postgresql.py
+++ b/storages/python/python-storage-plugins/postgresql.py
@@ -1,6 +1,7 @@
-"""PostgreSQL storage"""
+import os
+import sys
+import warnings
 import fnmatch
-from typing import TYPE_CHECKING
 
 import psycopg2
 from psycopg2 import sql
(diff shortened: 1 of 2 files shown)


CasperWA (commented):
This is a step backwards and negates a couple of hours of work for me due to the trickiness of this test.
All CI tests ran successfully after the changes I implemented, so I see no reason for a need of reversion.

If anything, the issues listed in the original post of this PR should be fixed. We shouldn't go backwards, only forwards.

Create an issue for it with a reproducible example and fix it in a subsequent PR.

PR #427, opened 2023-01-04
Ensure that UUIDs are NUL-terminated

jesper-friis:
C strings must be NUL-terminated.


CasperWA:
@jesper-friis if there is an associated issue with this PR, please close it.

PR #428, opened 2023-01-04
Included tools to the pip package

jesper-friis:
Closes #472

If installed in an environment, the user will most likely have to add LD_LIBRARY_PATH to the activate script.


python/setup.py
```diff
@@ -208,12 +208,18 @@ def build_extension(self, ext: CMakeExtension) -> None:
     # See issue #222: https://github.com/SINTEF/dlite/issues/222
     # extras_require={"all": extra_requirements},
     packages=["dlite"],
+    scripts=[
+        #str(SOURCE_DIR / "bindings" / "python" / "scripts" / "dlite-validate"),
+    ],
```
CasperWA, line 213:
```suggestion
```
If it's not used, it shouldn't be included.

If what you wish to do is to have callable scripts upon installation, use the `console_scripts` key in the `entry points` dictionary ([setuptools documentation](https://setuptools.pypa.io/en/latest/userguide/entry_point.html)).

python/setup.py
```diff
@@ -208,12 +208,18 @@ def build_extension(self, ext: CMakeExtension) -> None:
...
     package_data={
         "dlite": [
             dlite_compiled_ext,
             dlite_compiled_dll_suffix,
             str(Path(".") / "share" / "dlite" / "storage-plugins" /
                 dlite_compiled_dll_suffix),
+            str(Path(".") / "bin" / "dlite-getuuid"),
+            str(Path(".") / "bin" / "dlite-codegen"),
+            str(Path(".") / "bin" / "dlite-env"),
```
CasperWA, line 222: Could you not just do a single line with a glob expression?
E.g.,: `str(Path(".").resolve() / "bin" / "*")` or something similar.

CasperWA (changes requested):

python/setup.py
jesper-friis: It awaits PR #461 to be merged. When that is done, it should be uncommented.

python/setup.py
jesper-friis: No, that is dangerous and might pick up things we don't want to install.

CasperWA:
Have you tested this works as intended upon building and installing the Python package?

jesper-friis:
> Have you tested this works as intended upon building and installing the Python package?

How do we get access to the python package before it is made available on pypi?

CasperWA:
> > Have you tested this works as intended upon building and installing the Python package?
> 
> How do we get access to the python package before it is made available on pypi?

I would just build it locally, and possibly create a new virtualenv to install the built wheel?

PR #431, opened 2023-01-05
401 example 4 is not working properly

lovfall:
Implemented working example based on C example 1. 
This PR depends on https://github.com/SINTEF/dlite/pull/414

Clean up and documentation is still missing, this PR should not be merged before that is done


examples/ex4/CMakeLists.txt
```diff
@@ -2,36 +2,51 @@
...
   ${DLITE_ROOT}/cmake
   ${DLITE_ROOT}/cmake/Modules
   ${DLITE_ROOT}/share/cmake
   ${DLITE_ROOT}/share/cmake/Modules
   )
-include(Finddlite)
-find_package(dlite)
 
+
+set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${DLITE_ROOT}/share/dlite/cmake)
```
jesper-friis, line 23: This can be included in the `list(APPEND ...)` above

jesper-friis (approved):

jesper-friis (changes requested):
The github CI should be fixed

examples/ex4/CMakeLists.txt
lovfall: Have no clue why, but that does not work

jesper-friis: Ok, then we keep it

jesper-friis (approved):

PR #461, opened 2023-01-05
Added dlite-validate script

### LINKED ISSUE #353: Datamodel dimensions

jesper-friis:
Closes #353 

Added a simple script for validating data instances and metadata .


bindings/python/CMakeLists.txt
```diff
@@ -215,14 +215,21 @@ install(
...
-  )
+)
+install(
+  CODE "
+   execute_process(
+     COMMAND ${Python3_EXECUTABLE} setup.py install
+       --prefix=${CMAKE_INSTALL_PREFIX} --force
+     WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+   )"
+)
```
francescalb, line 232: I do not understand what this has to do with the issue.  It looks to me like this is installing dlite-python. Did we not do that before?

bindings/python/scripts/dlite-validate
```diff
@@ -0,0 +1,76 @@
...
+
+
+def main():
+    """Datamodel validator"""
+    parser = argparse.ArgumentParser(
+        description='DLite datamodel validator.',
+    )
+    parser.add_argument(
+        'url',
+        help='URL to storage to validate.  Relative file paths are allowed.'
```
francescalb, line 45: Wouldn't it be clearer to write "URL to datamodel to be validated. ..."

bindings/python/scripts/dlite-validate
```diff
@@ -0,0 +1,76 @@
...
+    )
+    parser.add_argument(
+        '--driver', '--plugin', '-d',
+        help='Storage plugin to use when loading the instance.  If `url` '
+        'is a file path, the default is to infer the driver from the file '
+        'extension.',
+    )
+    parser.add_argument(
+        '--options', '-o',
+        help='Options parsed on to the dlite storage plugin.'
```
francescalb, line 55: What kind of options are these?

bindings/python/scripts/dlite-validate
```diff
@@ -0,0 +1,76 @@
...
+import yaml
+from pathlib import Path
+
+import dlite
+
+
+def parse(url, driver, options=None):
+    """Parse url and return an instance."""
+    match = re.match(r'^([a-zA-Z][a-zA-Z0-9+.-]*)://', url)
+    if match and match.groups[0].lower() != 'file':
```
francescalb, line 16: There seems to be an error here: match.groups()
```suggestion
    if match and match.groups()[0].lower() != 'file':
```

bindings/python/scripts/dlite-validate
```diff
@@ -0,0 +1,76 @@
...
+import json
+import re
+import yaml
+from pathlib import Path
+
+import dlite
+
+
+def parse(url, driver, options=None):
+    """Parse url and return an instance."""
```
francescalb, line 14: Can we have a little longer docstring that also includes some examples of usage for the various options?

francescalb (changes requested):
This looks good, but there are no tests. Please add tests for possible outcomes and inputs.

bindings/python/CMakeLists.txt
jesper-friis: > I do not understand what this has to do with the issue. It looks to me like this is installing dlite-python. Did we not do that before?

This install command runs the new setup.py in the bindings/python/ directory. 

The rationale is that we use cmake for building and installing dlite. 
But setuptools knows better how to install python scripts, like how to add a shebang (#!/path/to/python/interpreter/for/current/virtualenv) on the first line. 
To achieve this, cmake calls `setup.py install` to install the scripts.

bindings/python/scripts/dlite-validate
jesper-friis: Improved docstring has now been added together with --id and --storage-path options for validating data instances.

bindings/python/scripts/dlite-validate
jesper-friis: Well spottet!

bindings/python/scripts/dlite-validate
jesper-friis: dlite.Instance.from_location() takes four arguments, driver, location, options and id. This flag make it possible to provide a value to the options argument. This is not needed for validating a datamodel stored in json or yaml files, but may be used to e.g. provide username/password for validating a datamodel stored in a postgresql database.

I updated the help text to: "Semicolon-separated list of options parsed on to the dlite storage plugin."
Should we include more of the above explanation?

bindings/python/CMakeLists.txt
francescalb: Thank you for the explanation.

bindings/python/scripts/dlite-validate
francescalb: I think at least one example would be good. Otherwise it is very cryptic what kind of options these really are.

jesper-friis: Ok, da har jeg lagt til et eksempel.

La også til et par ekstra tests med forenklet syntax via et par brukerdefinerte cmake-funksjoner.

jesper-friis:
> This looks good, but there are no tests. Please add tests for possible outcomes and inputs.

There should be 5 tests - all implemented in the CMakeLists.txt file in the new scripts/ directory.

francescalb (approved):

PR #467, opened 2023-01-09
Coding guidelines

jesper-friis:
This is an initial stub for a coding guideline. 

I don't think we should spend too much time on this, but it is good to have under the documentation for developments.

### DIFF
--- a/doc/coding_guideline.md
+++ b/doc/coding_guideline.md
@@ -22,7 +22,12 @@ If they due to a bug happen anyway, you rather want an early and clean abort, th
 Testing
 -------
 New features should be tested.
-Tests should be added to
+All tests are run with [ctest].
+
+Testing of C code is currently done with the minimal [MinUnit] C/C++ testing framework.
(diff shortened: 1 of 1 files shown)


daniel-sintef:
I think we should have a clearer sense of what does and does not go into one of these coding guidelines. Personally we should not try to write too much 'generic' advice, such as 'use assert statements'. Instead we should point to some good standard resource and only mention things that are *specific* to Dlite, e.g., use FAIL_CODE, or things that deviate from standard practice.

jesper-friis:
> I think we should have a clearer sense of what does and does not go into one of these coding guidelines. Personally we should not try to write too much 'generic' advice, such as 'use assert statements'. Instead we should point to some good standard resource and only mention things that are _specific_ to Dlite, e.g., use FAIL_CODE, or things that deviate from standard practice.

I agree, we shouldn't be too generic in such a document. But assert is rarely used in Python, so I thing it is worth to mention it and when it should be used and when to raise an exception instead.

PR #472, opened 2023-01-14
Extend the storage plugin api

### LINKED ISSUE #80: Simplify the storage plugin API

### LINKED ISSUE #335: API for accessing plugin documentation
Add an entry for a GetHelp() function pointer in `Struct _DLiteStoragePlugin` in `src/dlite-storage-plugins.h`. 

For plugins supporting it, it should simply return a help string for the plugin, documenting the plugin and its options. Add a help function for generating the option documentation from a DLiteOpt struct. 

For Python plugins the documentation should simply be extracted from the `__doc__` attribute of the plugin class.

Relates to issue #80.

jesper-friis:
Implements the new features suggested in issues #80 and #335.

The storage plugin API has been extended with the following optional methods:
- Flush : Storage method that flushes cached data to the storage.
- Help: Return documentation of storage plugin.
- DeleteInstance: Storage method that deletes an instance in the storage. Now storage plugins supports full CRUD.
- MemLoadInstance: Driver/class method that loads an instance from a memory buffer (instead of the storage).
- MemSaveInstance: Driver/class method that saves an instance to a memory buffer (instead of the storage).

In order to close #80, we should update the hdf5 storage plugin to use the new API and remove the old deprecated API. 
That should probably be a separate PR.

Closes #335


daniel-sintef:
I don't understand everything. But with the new tests I am OK with accepting the pull request.

daniel-sintef (approved):

PR #478, opened 2023-01-30
More homogen icons for the manual

jesper-friis:
I picked some icons from https://www.svgrepo.com/collections/ and colored them SINTEF-blue to get a more homogen look.

Does sphinx support svg or should the icons be converted to png?


CasperWA:
> Does sphinx support svg or should the icons be converted to png?

You should be able to get the answer to this by building and viewing the documentation locally.

lovfall (approved):
Looks great!

PR #481, opened 2023-01-31
Specify explicitely that we use fno to represent mapping functions

jesper-friis:
Specify explicitely that we use fno to represent mapping functions in the mappings example.


daniel-sintef (approved):
Looks good. Can you comment briefly here why this is needed though?

PR #483, opened 2023-02-01
466 describe how to add ld library path to virtualenv activate scrip

alfredoisg:


jesper-friis (approved):
Very nice with a good explanation

alfredoisg:
In the last demo meeting, Francesca suggested that the note could be moved to the section **Setting up the environment**. It actually makes sense. How do I proceed? Should I restore the branch?

alfredoisg:
@jesper-friis

PR #485, opened 2023-02-03
Mongodbatlas dlite

daniel-sintef:
Ignore the unit tests for now. See my ideas for changing the dlite interface


daniel-sintef:
From our discussion today client_options should be merged into options

doc/contributors_guide/storage_plugins_mongodb-atlas.md
```diff
@@ -0,0 +1,33 @@
+Mongodb Storage plugin
```
jesper-friis, line 1: Great documentation, but I think it belongs to the user_guide.

storages/python/python-storage-plugins/mongodb.py
```diff
@@ -11,7 +11,7 @@
 
 class mongodb(dlite.DLiteStorageBase):
     """DLite storage plugin for PostgreSQL."""
-    def open(self, uri, options=None):
+    def open(self, uri, options=None, client_options=None):
```
jesper-friis, line 14: The `open()` method should only take `uri` and `options` as arguments. I am afraid we have to encode the client_options in `options`.

doc/contributors_guide/storage_plugins_mongodb-atlas.md
```diff
@@ -0,0 +1,33 @@
...
+-------
+
+This is a quickguide for users wishing to connect to a mongodb atlas server. It is assumed that they have access 
+to a server and have their own username and password. 
+
+```python
+    >>> user = "USERNAME"
+    >>> password = "PASSWORD"
+    >>> server = "MYSERVERADDRESS"
+    >>> uri = f"mongodb+srv://{server}"
```
jesper-friis, line 14: We have to think how to handle the "mongodb+srv" scheme correctly. The issue is that the URL passed to functions like `dlite.Instance.from_url()` is split up internally by dlite into four components

    <SCHEME>://<LOCATION>?<OPTIONS>#<ID>

where the <SCHEME> part is used to identify the plugin. But we cannot create a plugin named `mongodb+srv` since it is not a valid Python identifier. I don't think this is an issue for the plugins written in C, since the plugin name is never used a s an identifier in that case. 

I am not sure what would be the best solution. One possibility could be that the C API that loads the Python plugins, strips off the "+" and everything that follows it, when looking for a driver and then provides the full scheme as an optional argument.

jesper-friis (changes requested):

storages/python/tests-python/test_mongodb-atlas_python.py
```diff
@@ -0,0 +1,41 @@
+"""Script to test the 'mongodb' DLite plugin from Python."""
```
jesper-friis, line 1: This script needs a little cleanup, but we can do that when we have a good and working solution.

A think it would be good to have two tests, one that uses the anonymous user which should always run, and one that require a username and password. 

We should also document a generic way to add the `user_password_file`.

storages/python/python-storage-plugins/mongodb.py
```diff
@@ -22,23 +22,30 @@ def open(self, uri, options=None):
...
         opts = Options(
             options,
-            defaults='database=dlite;collection=dlite_coll;mode=w;'
-            'authMechanism=SCRAM-SHA-256;mock=no',
+            defaults='database=dlite;collection=dlite_coll;mode=w;mock=no',
         )
         opts.setdefault('password', None)
         self.writable = True if 'w' in opts.mode else False
 
+        client_options = {}
```
jesper-friis, line 46: This line can be removed

jesper-friis, line 46:
```suggestion
```
This line can be removed.

doc/contributors_guide/storage_plugins_mongodb-atlas.md
daniel-sintef: I have made a solution that should be handle all of these cases, see the latest commit

storages/python/tests-python/test_mongodb-atlas_python.py
daniel-sintef: The test has been modified to take all the relevant information from environment variables, not sure if we need to document this as it is a standard technique. Our CI would then need to be updated to export these environment variables at run time.

storages/python/tests-python/test_mongodb-atlas_python.py
```diff
@@ -0,0 +1,31 @@
+import sys
+import os
+from pathlib import Path
+import dlite
+
+# Load user and password from environment variables
+user = os.environ.get("DLITETEST_MONGODB_USER")
+password = os.environ.get("DLITETEST_MONGODB_PASSWORD")
+server = os.environ.get("DLITETEST_MONGODB_SERVER")
+inputdir = os.environ.get("DLITETEST_MONGODB_INPUTDIR")
```
jesper-friis, line 10: Setting the credentials in the environment is one way of providing them. But isn't it be more convenient to ask the user to add a local file with the credentials? That way, if you once have set up the credentials, it will run whenever you run the tests, without any need to set up the environment.

The file with credentials should of cause be listed in .gitignore to avoid that it by mistake is committed.

This is how we do in the C tests for postgresql.

jesper-friis (commented):
The changes looks fine, but it would be nice to be able to run the tests.

storages/python/tests-python/test_mongodb-atlas_python.py
daniel-sintef: The test is now robust and will try to use either environment variables or a file.

daniel-sintef:
Added https://github.com/SINTEF/dlite/issues/512

jesper-friis (approved):

PR #505, opened 2023-03-21
Update yaml plugin

jesper-friis:
# Description:
Fixed some bugs in the yaml plugin:
- Don't dublicate UUID in YAML output.
- Update default to not overwrite yaml files (in single-entity form) that contains an entity.

## Type of change:
- [x] Bug fix.
- [ ] New feature.
- [ ] Documentation update.


storages/python/python-storage-plugins/yaml.py
```diff
@@ -89,7 +93,8 @@ def save(self, inst: dlite.Instance):
 
         """
         self._data[inst.uuid] = inst.asdict(
-            soft7=dlite.asbool(self.options.soft7)
+            soft7=dlite.asbool(self.options.soft7),
+            uuid=self.single,
```
CasperWA, line 97: Not sure what this means. But I'm guessing it means that only a single instance from `inst` is returned? Or multiple?

storages/python/python-storage-plugins/yaml.py
```diff
@@ -74,12 +73,17 @@ def load(self, id: str):
...
             self._data,
             id,
             single=self.options.single,
             check_storages=False,
         )
+        # Metadata in single-entity form is always read-only
+        if inst.is_meta and self.single:
+            self.writable = False
+
+        return inst
```
CasperWA, line 86: Is this ensuring the instance cannot be overwritten? This confuses me a bit. Here we've already loaded the instance, right? So setting the `writable` property afterwards - does that change anything?

storages/python/python-storage-plugins/yaml.py
jesper-friis: It means that it is only when we we to a yaml file in single-entity form that we want to include the UUID. In multiple-entity form, the UUID (or possible the URI) will already be provided as the key for the entity.

storages/python/python-storage-plugins/yaml.py
jesper-friis: Yes, it makes a difference. It is not the instance we want to write-protect, but the yaml file. 

It is first after we have loaded the instance we know whether it is a metadata in single-entity form.

storages/python/python-storage-plugins/yaml.py
```diff
@@ -74,12 +73,17 @@ def load(self, id: str):
...
             A DLite Instance corresponding to the given `id` (UUID).
         """
-        return instance_from_dict(
+        inst = instance_from_dict(
             self._data,
             id,
             single=self.options.single,
             check_storages=False,
         )
+        # Metadata in single-entity form is always read-only
```
francescalb, line 82:
```suggestion
        # Ensure metadata in single-entity form is always read-only
```

francescalb (approved):
Approved with one comment.

Issue #509, opened 2023-03-23
CI tests for example ex4 fails  


jesper-friis:
Double-check what is happening when we run ex4 as a unit test. It works fine locally, but fails on GitHub. If you rerun the test on GitHub one or two times it sometimes succeeds (?) 

CI/CD output shown on GitHub when ex4 fails is shown below.

**Edit:** The problem is that cmake finds, but does not accept the installed `dliteConfig.cmake`. Unfortunately, it provides not explanation of why it doesn't accept the file.
This is only an issue that shows up with CMake 3.26.0. Everything works fine with cmake 3.24.2 and 3.25.2.
Unfortunately the ubuntu now comes with CMake 3.26.0 in the `ci_tests.yml` workflow.

```
The following tests FAILED:
	 94 - ex4-config (Failed)
	 95 - ex4-build (Failed)
	 96 - ex4-test (Failed)
Errors while running CTest
Output from these tests are in: /home/runner/work/dlite/dlite/build/Testing/Temporary/LastTest.log
Use "--rerun-failed --output-on-failure" to re-run the failed cases verbosely.
UpdateCTestConfiguration  from :/home/runner/work/dlite/dlite/build/DartConfiguration.tcl
Parse Config file:/home/runner/work/dlite/dlite/build/DartConfiguration.tcl
UpdateCTestConfiguration  from :/home/runner/work/dlite/dlite/build/DartConfiguration.tcl
Parse Config file:/home/runner/work/dlite/dlite/build/DartConfiguration.tcl
Test project /home/runner/work/dlite/dlite/build
Constructing a list of tests
Done constructing a list of tests
Updating test list for fixtures
Added 0 tests to meet fixture requirements
Checking test dependency graph...
Checking test dependency graph end
test 94
    Start 94: ex4-config

94: Test command: /usr/local/bin/cmake "-G" "Unix Makefiles" "-DCMAKE_BUILD_TYPE=Debug" "-DCMAKE_INSTALL_PREFIX=/home/runner/.local" "-DCMAKE_C_FLAGS=-Wno-missing-field-initializers -m64 -Wall -Wextra -Wpedantic" "-DDLITE_ROOT=/home/runner/.local" "/home/runner/work/dlite/dlite/examples/ex4"
94: Working Directory: /home/runner/work/dlite/dlite/build/examples/ex4
94: Test timeout computed to be: 1500
94: CMake Error at CMakeLists.txt:5 (find_package):
[...]

PR #510, opened 2023-03-23
Updated ci-tests.yml to use cmake 3.25.2

jesper-friis:
# Description:
Fix failing CI tests by downgrading cmake from 3.26.0 to cmake 3.25.2. This solves the issue of calling `find_package(dlite)` when running example ex4 as a CI test  - see issue #509.

This is only a temporary fix:
- If the issue is due to a bug in CMake 3.26.0, then this fix should be reverted when Ubuntu has updated CMake to a working version.
- If the issue is due to a bug in how we generate `dliteConfig.cmake` from `cmake/dliteConfig.cmake.in`, this fix should be reverted when we have fixed that problem.

## Type of change:
- [x] Temporary workaround.
- [ ] Bug fix.
- [ ] New feature.
- [ ] Documentation update.

### DIFF
--- a/.github/workflows/ci_tests.yml
+++ b/.github/workflows/ci_tests.yml
@@ -14,7 +14,7 @@ jobs:
         sudo apt-get update --fix-missing
 
     - name: Workaround - install CMake 3.25.2 since 3.26.0 doesn't work
-      run |
+      run: |
         sudo apt remove cmake
         sudo apt purge --auto-remove cmake
         wget http://www.cmake.org/files/v3.25/cmake-3.25.2.tar.gz
--- a/.github/workflows/ci_tests.yml
+++ b/.github/workflows/ci_tests.yml
@@ -9,16 +9,29 @@ jobs:
     steps:
     - uses: actions/checkout@v3
 
-    - name: Update submodules
-      run: |
-        git submodule init
-        git submodule update
-    - name: Get Cmake version
+    - name: apt-get update
+      run |
+        sudo apt-get update --fix-missing
+
+    - name: Workaround - install CMake 3.25.2 since 3.26.0 doesn't work
+      run |
+        sudo apt remove cmake
+        sudo apt purge --auto-remove cmake
+        wget http://www.cmake.org/files/v3.25/cmake-3.25.2.tar.gz
+        tar xf cmake-3.25.2.tar.gz
+        cd cmake-3.25.2
+        ./configure
+        make
+        sudo make install
+        hash -r
+        cd -
+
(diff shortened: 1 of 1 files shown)


.github/workflows/ci_tests.yml
```diff
@@ -9,16 +9,29 @@ jobs:
...
-      run: |
-        git submodule init
-        git submodule update
-    - name: Get Cmake version
+    - name: apt-get update
+      run |
+        sudo apt-get update --fix-missing
+
+    - name: Workaround - install CMake 3.25.2 since 3.26.0 doesn't work
+      run |
```
CasperWA, line 17:
```suggestion
      run: |
```

.github/workflows/ci_tests.yml
```diff
@@ -9,16 +9,29 @@ jobs:
...
     - uses: actions/checkout@v3
 
-    - name: Update submodules
-      run: |
-        git submodule init
-        git submodule update
-    - name: Get Cmake version
+    - name: apt-get update
+      run |
+        sudo apt-get update --fix-missing
```
CasperWA, line 14:
```suggestion
      run: sudo apt-get update --fix-missing
```

.github/workflows/ci_tests.yml
```diff
@@ -9,16 +9,29 @@ jobs:
...
+      run |
+        sudo apt remove cmake
+        sudo apt purge --auto-remove cmake
+        wget http://www.cmake.org/files/v3.25/cmake-3.25.2.tar.gz
+        tar xf cmake-3.25.2.tar.gz
+        cd cmake-3.25.2
+        ./configure
+        make
+        sudo make install
+        hash -r
```
CasperWA, line 26: Just out of curiosity, why are you running this? Is it necessary for cmake? I suppose it is?

CasperWA (changes requested):
I've noted any optional suggested changes as such. I.e., if suggested changes are not noted as such, they are considered "required" from my point-of-view.

There are also a few comments of interest.

.github/workflows/ci_tests.yml
jesper-friis: Yes, we have to build cmake from source - at least I didn't found historical .dep packages for cmake.

CasperWA:
@jesper-friis I would've liked to do another review on this before merging, after having resolved all discussions....

.github/workflows/ci_tests.yml
CasperWA: This was a _required_ change due to the missing colon (`:`)!

PR #511, opened 2023-03-23
Added "integer" as a valid type name

jesper-friis:
# Description:
Adding "integer" as an alias for "int64". Motivated by Otello writing "integer" in his entities, which shows that it is a natural alias to use for newcomers to dlite.

**Question**: what should be the precision of unspecified types like "integer", "int" or "float"? 
- If we are C-oriented, the natural answer would be `sizeof(int)` for "int" and `sizeof(float)` for "float". Both of these will correspond to 32 bits on both X86 and x64_86 platforms. 
- If we are Python-oriented, the natural answer would be 64 bits for all of them.

The current implementation is C-oriented for the precision of "int" and "float" types. Is that the best choice? 

This PR is defines "integer" to 64 bit. Does that make sense?

## Type of change:

- [ ] Bug fix.
- [x] New feature.
- [ ] Documentation update.

- [x] Is the change limited to one issue?
- [x] Does this PR close the issue?
- [x] Is the code easy to read and understand?


lovfall:
The size of int in C is technically system dependent, and is defined in stdint.h, but I guess for most systems this is 32bit. Given that the size of int(eger) might not be know by the end user, the safe bet would be to go 64 bit. So I agree with your conclusion

lovfall (approved):

lovfall:
Is there some documentation to update?

Issue #515, opened 2023-03-25
Instantiate properties from dict


jesper-friis:
Make it possible to instantiate properties from dict. An important use case is to make it easier to work with dicts, dataclasses and pydantic datamodels. We should make it easy to convert to/from nested dicts and instances with ref-type properties.

This could be exposed via:
- A `dlite.Instance.set_dict(dict_obj)` method that sets properties of an instance from a dict.
- A `dlite.Instance.get_dict()` method that returns the properties as a dict.
- Metadata instantiation from nested dicts, like
  ```python
  Entity = dlite.get_instance("http://onto-ns.com/meta/0.1/Entity")
  Entity(dimensions={...}, properties={"ref": {...}, other_prop=True, ...})
  ```

At C-level we could add a function

```C
```

Issue #501, opened 2023-03-15
Have a proper pydantic -> dlite entity function


CasperWA:
It would be straightforward to have a small utility function to generate DLite entities (data models) based off of pydantic models (classes) or instances.
The current functionality in `dlite.utils.to_metadata()` expects a pydantic instance and of a specific type as well, as the function passes on the `dict`-serialized pydantic instance to `dlite.utils.instance_from_dict()`, which expects the given `dict` to already have the correct data model structure.

jesper-friis:
Do you have an example of a pydantic datamodel that you would like to generate an entity for?

CasperWA:
This can be any data model.
For reference you could checkout the pydantic models in [OTEAPI Core (here the transformation config)](https://github.com/EMMC-ASBL/oteapi-core/blob/master/oteapi/models/transformationconfig.py) or in [OPTIMADE Python tools (here a replication of JSON:API v1)](https://github.com/Materials-Consortia/optimade-python-tools/blob/master/optimade/models/jsonapi.py).

CasperWA:
Essentially, there are several ways to do this:

1. Go through the OpenAPI schema (a JSON file, which all pydantic models can generate).
2. Go through the Python API directly, using either pydantic classes or instances.

PR #516, opened 2023-03-26
501 pydantic dlite entity function

jesper-friis:
# Description:
Addresses issue #501 

This is a quick and rough implementation and demonstration of how to generate (i) dlite metadata from a pydantic model and (ii) a dlite instance from it. Supports nested pydantic models, but not all pydantic features are supported. Include a new dedicated test. Code is not pretty...

Handling of nested structures would have been easier and cleaner with issue #515 in place. 

## Type of change:

- [x] Bug fix.
- [x] New feature.
- [ ] Documentation update.


jesper-friis:
The failing CI test for windows is probably because it somehow installs dlite-python from an outdated cache. I have not digged into it...

bindings/python/tests/test_pydantic.py
```diff
@@ -0,0 +1,77 @@
...
+
+# Skip test if pydantic isn't installed
+try:
+    from pydantic import BaseModel, Field
+except ImportError:
+    import sys
+    sys.exit(44)
+
+
+class TransformationStatus(BaseModel):
```
kriwiik, line 15: This is not used in the test and can be removed

bindings/python/tests/test_pydantic.py
```diff
@@ -0,0 +1,77 @@
...
+    banana = 'y'
+
+
+class Spam(BaseModel):
+    foo: Foo
+    bars: List[Bar]
+
+m = Spam(foo={'count': 4}, bars=[{'apple': 'x1'}, {'apple': 'x2'}])
+
+from dlite.utils import pydantic_to_metadata, pydantic_to_property
```
kriwiik, line 68: pydantic_to_metadata is already imported on line 5. Move the import of pydantic_to_property to line 5

bindings/python/tests/test_pydantic.py
```diff
@@ -0,0 +1,77 @@
+from datetime import datetime
+from typing import List, Optional
+
+import dlite
```
kriwiik, line 4: Not  accessed

bindings/python/dlite-entity-python.i
```diff
@@ -158,11 +158,12 @@ def get_instance(id: "str", metaid: "str" = None, check_storages: "bool" = True)
 %extend _DLiteProperty {
   %pythoncode %{
     def __repr__(self):
+        ref = ', ref=%r' % self.ref if self.ref else ''
         shape = ', shape=%r' % self.shape.tolist() if self.ndims else ''
         unit = ', unit=%r' % self.unit if self.unit else ''
         descr = ', description=%r' %self.description if self.description else ''
-        return 'Property(%r, type=%r%s%s%s)' % (
-            self.name, self.type, shape, unit, descr)
+        return 'Property(%r, type=%r%s%s%s%s)' % (
```
kriwiik, line 165: How does this work? If I understand it correctly,  everything except self.name passed to the type argument. Why?

bindings/python/dlite-python.i
```diff
@@ -874,24 +876,28 @@ int dlite_swig_set_scalar(void *ptr, DLiteType type, size_t size, obj_t *obj)
         dest->description = (src->description) ? strdup(src->description) :NULL;
 
       } else if (PySequence_Check(obj) && PySequence_Length(obj) == 6) {
+        /* We should deprecate this - either remove or change to dict... */
```
kriwiik, line 879: Why?

bindings/python/dlite-entity-python.i
jesper-friis: Since all arguments, except `name` and `type` are optional, we wither want to include them in the returned `repr()` representation as strings like 

    ", shape=['N', 'N+1']" 

if they are set or the empty string "" if they are unset (scalar, in case of shape). 

Are the variable names (`ref`, `shape`, `unit` and `descr`) are misleasing?

bindings/python/dlite-python.i
jesper-friis: Because I feel it is too easy to make unnoticed mistakes if we allow to set a property sequence of the form:

```python
(name, type, ref, shape, unit, descr)
```

It will also be susepatble to errors if we in the future make any changes to the property arguments.

It think it will be more explicit and robust to instead require a dict of the form

```python
{"name": name, "type": type, "ref": ref, "unit": unit, "description": descr}
```

bindings/python/tests/test_pydantic.py
jesper-friis: Well spotted. Removed

bindings/python/tests/test_pydantic.py
jesper-friis: The `TransformationStatus` class is used to instantiate `t`. I improved the test with additional assert statements.

bindings/python/tests/test_pydantic.py
jesper-friis: ... readded again because it is needed after I improved the tests.

kriwiik (approved):

PR #522, opened 2023-03-31
Increased the size of types with unspecified size.

jesper-friis:
The motivation for this change is that it is better to be safe than sorry. This is also in accordance with feedback on review on PR #511. If the user know that the number can be represented with fewer bytes, he/she can always use one of the explicit types, like "int32".

- Increaded the size of "int" and "uint" to 8 bytes.
- Increased the size of "float" and `sizeof(double)` bytes.

Now "integer" is simply an alias for "int", which seems very reasonable.

The danger that this PR should corrupt existing data is very small, since it is only about the number of bits dlite uses for its internal representation or a property. DLite never dump its internal data representation to disk, it is always serialised into a well-defined representation by the storage plugins. 

# Description:

## Type of change:

- [ ] Bug fix.
- [ ] New feature.
- [ ] Documentation update.

- [x] Is the change limited to one issue?
- [x] Does this PR close the issue?
- [x] Is the code easy to read and understand?
- [x] Do all new feature have an accompanying new test?
- [x] Has the documentation been updated as necessary?


lovfall (approved):

lovfall:
Don't see why one test fails, but should be investigated.

PR #562, opened 2023-06-24
Convince redis to not hang

jesper-friis:
Description
 ===========
Added `socket_keepalive` option and enabled it by default - this seems to counteract that the redis client hangs indefinitely.

Type of change
--------------
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Test update

Checklist for the reviewer
--------------------------

### DIFF
--- a/storages/python/python-storage-plugins/redis.py
+++ b/storages/python/python-storage-plugins/redis.py
@@ -17,7 +17,10 @@ class redis(dlite.DLiteStorageBase):
             - `port`: Port to connect to (default: 6379).
             - `username`: Redis user name.
             - `password`: Redis password.
-            - `ssl`: Whether to ssl-encrypt the connection (default: false).
+            - `socket_keepalive`: Whether to enable socket keepalive option.
+              Seems to protect against hanging.  Default: true
+            - `socket_timeout`: Timeout after given number of seconds.
+            - `ssl`: Whether to ssl-encrypt the connection.
             - `ssl_certfile`: Path to SSL certificate signing request (.crt).
             - `ssl_keyfile`: Path to SSL private key file (.key).
             - `ssl_ca_certs`: Path to SSL certificate container (.pem).
@@ -31,7 +34,7 @@ class redis(dlite.DLiteStorageBase):
               Generate the key with `crystallography.fernet.generate_key()`.
         """
         self.uri = uri
-        opts = Options(options, defaults="port=6379;ssl=false;db=0")
+        opts = Options(options, defaults="port=6379;socket_keepalive=true;db=0")
 
         # Pop out options passed to redis.set()
         self.setopts = {
@@ -45,7 +48,11 @@ class redis(dlite.DLiteStorageBase):
 
         # Fix option types and open Redis connection
         opts.port = int(opts.port)
-        opts.ssl = dlite.asbool(opts.ssl)
+        opts.socket_keepalive = dlite.asbool(opts.socket_keepalive)
+        if "socket_timeout" in opts:
+            opts.socket_timeout = int(opts.socket_timeout)
+        if "ssl" in opts:
+            opts.ssl = dlite.asbool(opts.ssl)
         opts.db = int(opts.db)
         self.redis = Redis(host=uri, **opts)
         self.closed = False
--- a/storages/python/tests-python/test_redis.py
+++ b/storages/python/tests-python/test_redis.py
@@ -27,7 +27,7 @@ save2 = inst2.asdict()
 
 
 # Test storing to Redis
-with dlite.Storage("redis", "localhost", "port=6379;expire=1;db=13") as s:
+with dlite.Storage("redis", "localhost", "expire=1;db=13") as s:
     try:
         s.save(inst1)
     except dlite.DLiteError as exc:


storages/python/python-storage-plugins/redis.py
```diff
@@ -31,7 +34,7 @@ def open(self, uri: str, options=None):
               Generate the key with `crystallography.fernet.generate_key()`.
         """
         self.uri = uri
-        opts = Options(options, defaults="port=6379;ssl=false;db=0")
+        opts = Options(options, defaults="port=6379;socket_keepalive=true;db=0")
```
Treesarj, line 37: Is the point to always maintain connection? is that a good idea?

jesper-friis: Not really. It is more practical. The point is that I had some issues with Redis was hanging for some reasons when I was testing it. Googling the issue suggested to set `socket_keepalive=true`. Trying it solved my problem, so I simply add it as a default.

Treesarj: but wouldn't it still hang when there are concurrent connections from multiple clients?- I am thinking from a deployment perspective and wouldn't it consume more resources and increase traffic because of keep-alive packets?

ajeklund: I would expect redis to be able to handle many simultaneous connections, so using keep-alive should not be a problem. I looked for the time interval for the keep-alive messages but wasn't able to find it. I expect it to be on the second or minute timescale, so that shouldn't cause traffic load problems, either.

storages/python/python-storage-plugins/redis.py
```diff
@@ -17,7 +17,10 @@ def open(self, location: str, options=None):
             - `port`: Port to connect to (default: 6379).
             - `username`: Redis user name.
             - `password`: Redis password.
-            - `ssl`: Whether to ssl-encrypt the connection (default: false).
+            - `socket_keepalive`: Whether to enable socket keepalive option.
+              Seems to protect against hanging.  Default: true
+            - `socket_timeout`: Timeout after given number of seconds.
+            - `ssl`: Whether to ssl-encrypt the connection.
```
ajeklund, line 23: Changing the ssl configuration was not mentioned in the commit messages and is not described elsewhere. Was this necessary for fixing the "redis hanging" problem?

Previously, it was clear that the `ssl` setting was `false` by default. Now, the default behavior is undefined as far as I can tell. Is this the intention?

storages/python/python-storage-plugins/redis.py
```diff
@@ -44,7 +47,11 @@ def open(self, location: str, options=None):
 
         # Fix option types and open Redis connection
         opts.port = int(opts.port)
-        opts.ssl = dlite.asbool(opts.ssl)
+        opts.socket_keepalive = dlite.asbool(opts.socket_keepalive)
+        if "socket_timeout" in opts:
```
ajeklund, line 51: Do you have to supply the `socket_timeout` parameter in order for redit not to hang? In that case, a typical working value would be helpful in the parameter description.

Do you know if this value is used by the keep-alive functionality, or if `socket_timeout` only configures other behavior?

storages/python/python-storage-plugins/redis.py
jesper-friis: I agree that it is strange to have to provide `keep-alive`. But without this option, running the tests hangs on my machine. It might very well be a fix for a symptom rather than the cause. But setting `keep-alive` was the suggestion I found when googling for redis hanging - and it actually works for me.  

Would a compromise be to accept this PR for now, but add an issue to investigate the real cause of the hanging problem?

storages/python/tests-python/test_redis.py
```diff
@@ -27,7 +27,7 @@
 
 
 # Test storing to Redis
-with dlite.Storage("redis", "localhost", "port=6379;expire=1;db=13") as s:
+with dlite.Storage("redis", "localhost", "expire=1;db=13") as s:
```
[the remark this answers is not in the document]
jesper-friis: It is not needed - redis already defaults to port 6379.

But I have no strong feelings about this. I actually think I removed it just to check whether the port number is needed and didn't bother to re-add it again.

storages/python/python-storage-plugins/redis.py
ajeklund: I do not have a problem with using `keep-alive`, since it increase robustness. It can surely reduce problems that are related to the network (and its policies) between the client and the server.

The PR is not linked to an issue, and the PR description does not describe whether the problems experienced with redis occur when redis is on a remote or local computer. Can you please clarify this?

storages/python/python-storage-plugins/redis.py
jesper-friis: It was a part of my testing around for a solution for the hanging problem - but has nothing to do with the hanging problem.

However, I kept ssl unset, since I think it is a bad habit to turn off security by default. Now we leave it to redis to select the default value. I believe they have a better feeling for a sensible default value than I have.

storages/python/tests-python/test_redis.py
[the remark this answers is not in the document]
ajeklund: OK, then it makes sense.

storages/python/python-storage-plugins/redis.py
jesper-friis: The problem is as far as I can say not consistent. I have from time to time seen it on GitHub CI/CD. During the initial development of the plugin I had no problem with redis hanging, but then it started while I was working on unrelated things.

jesper-friis: > but wouldn't it still hang when there are concurrent connections from multiple clients?- I am thinking from a deployment perspective and wouldn't it consume more resources and increase traffic because of keep-alive packets?

You might very well be right that setting keep-alive might not be optimal in deployment. But this PR just change the default for robustness. It is easy to set keep-alive=false in production.

storages/python/python-storage-plugins/redis.py
ajeklund: OK, so the security change is intentional. For traceability it would have been better to do this in a dedicated PR, but let's keep it in this one anyway.

storages/python/python-storage-plugins/redis.py
jesper-friis: Again, adding `socket_timeout` was a part of my testing for a solution for the hanging. It turned out not to help. I kept it as an option for the plugin in case a user of DLite wants to play with. The default is left to redis. 
I think that a good working value depends on the situation - redis is fast, so I guess it could be a lot less than a second.

storages/python/python-storage-plugins/redis.py
ajeklund: I don't see any problems with keep-alive. Those packages will be sent infrequently and only when there is low activity on the connection. There is no need to look further into this at this point.

storages/python/python-storage-plugins/redis.py
ajeklund: Thank you for the clarification!

ajeklund (approved):
Thank you for fixing this!

Issue #569, opened 2023-07-05
storages/python/python-storage-plugins/http.py needs requests


francescalb:
Two errors:
requests is needed in requirements.

A test should be made for this plugin.

jesper-friis:
Regarding requirements, I would prefer to add `requests` to a new `requirements_full.txt` file (see issue #517). 

A simple test using github is added in PR #593. The plugin was initially implemented to support the [entity-service](https://github.com/SINTEF/dlite-entities-service), but that seems never to get up and running...

PR #593, opened 2023-08-13
Test http plugin

jesper-friis:
# Description
Added a test for the http plugin (issue #569).

Also cleaned up the plugin and improved documentation.

## Type of change
- [ ] Bug fix
- [x] New feature
- [x] Documentation update
- [ ] Test update

## Checklist for the reviewer

### DIFF
--- a/storages/python/tests-python/test_http.py
+++ b/storages/python/tests-python/test_http.py
@@ -1,0 +1,13 @@
+try:
+    import requests
+except ModuleNotFoundError:
+    print("requests not installed, skipping test")
+    raise SystemExit(44)  # skip test
+
+import dlite
+
+
+url = "https://raw.githubusercontent.com/SINTEF/dlite/master/storages/python/tests-python/input/test_meta.json"
+
+meta = dlite.Instance.from_location("http", url)
+print(meta)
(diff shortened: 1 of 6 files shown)


storages/python/tests-python/test_http.py
```diff
@@ -0,0 +1,13 @@
+try:
+    import requests
+except ModuleNotFoundError:
```
francescalb, line 3: I don't like this since it makes it very easy to miss out if things are working or not in ci/cd. Our ci/cd workflows should install all requirements and test everything.

It is hard to see the output of the workflow here, it just passes, but it seems to me that requests is still not added to the requirements, which means that the test is passing without having even tried and it should fail.

francescalb (changes requested):
Please check my comments.

storages/python/tests-python/test_http.py
jesper-friis: I double-checked this. If I uninstall requests, ctest will report test_http as skipped, not as passed.

I also added requests to requirements.txt, so that we ensure that test_http is run by CI/CD on github.

storages/python/tests-python/test_http.py
```diff
@@ -0,0 +1,13 @@
...
+    print("requests not installed, skipping test")
+    raise SystemExit(44)  # skip test
+
+import dlite
+
+
+url = "https://raw.githubusercontent.com/SINTEF/dlite/master/storages/python/tests-python/input/test_meta.json"
+
+meta = dlite.Instance.from_location("http", url)
+print(meta)
```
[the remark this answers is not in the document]
jesper-friis: Ok, added an assert statement to also check the result in addition to that Instance.from_location() doesn't through an exception...

francescalb (approved):
Very good

PR #633, opened 2023-09-01
Support quoted relations in collection

jesper-friis:
# Description
We use the collection as a local knowledge base. Since n3-represented literals contains embedded double-quotes it is essential that serialisation and de-serialisation of collections retain embedded double-quotes.

This PR ensures that embedded double-quotes are escaped in the internal representation of relations, which is used by the built-in json encoder/decoder.

## Type of change
- [x] Bug fix & code cleanup
- [x] New feature
- [ ] Documentation update
- [x] Test update

## Checklist for the reviewer

### DIFF
--- a/src/dlite-type.c
+++ b/src/dlite-type.c
@@ -863,7 +863,13 @@ int dlite_type_print(char *dest, size_t n, const void *p, DLiteType dtype,
   case dliteRelation:
     {
       DLiteRelation *r = (DLiteRelation *)p;
-      m = snprintf(dest, n, "[\"%s\", \"%s\", \"%s\"]", r->s, r->p, r->o);
+      m = snprintf(dest, n, "[");
+      m += strquote(dest+m, PDIFF(n, m), r->s);
+      m += snprintf(dest+m, PDIFF(n, m), ", ");
+      m += strquote(dest+m, PDIFF(n, m), r->p);
+      m += snprintf(dest+m, PDIFF(n, m), ", ");
+      m += strquote(dest+m, PDIFF(n, m), r->o);
+      m += snprintf(dest+m, PDIFF(n, m), "]");
     }
     break;
   }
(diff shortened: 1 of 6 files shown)


src/dlite-type.c
```diff
@@ -1198,23 +1204,50 @@ int dlite_type_scan(const char *src, int len, void *p, DLiteType dtype,
         return errx(dliteParseError, "relation should have 3 (optionally 4) elements");
       m = tokens->end - tokens->start;
       if (tokens->type == JSMN_ARRAY) {
+        size_t bufsize=0;
+        char *buf=NULL;
         if (!(t = jsmn_element(src, tokens, 0))) return -1;
-        rel->s = strndup(src + t->start, t->end - t->start);
+        if (strnput_unquote(&buf, &bufsize, 0, src + t->start,
```
quaat, line 1210: seems to be a lot of hard-to-read copy-pasted code here. Please create a function, or even a macro to simplify the reading.

src/utils/strutils.c
```diff
@@ -268,14 +271,19 @@ int strnquote(char *dest, size_t size, const char *s, int n,
     if (size > i) dest[i] = '"';
     i++;
   }
-  while (s[j] && (n < 0 || (int)j < n)) {
-    if (s[j] == '"' && !(flags & strquoteNoEscape)) {
-      if (size > i) dest[i] = '\\';
+  if (s) {
```
quaat, line 274: hard to follow the logic. Comment line-by-line what is happening here, or simplify the expression with standard functions.

src/utils/strutils.c
```diff
@@ -342,6 +350,42 @@ int strnunquote(char *dest, size_t size, const char *s, int n,
 }
 
 
+/*
+  Like strnunquote(), but reallocates the destination and writes to
+  position `pos`.
+
+  On allocation error, -3 is returned.
+ */
+int strnput_unquote(char **destp, size_t *sizep, size_t pos, const char *s,
```
quaat, line 359: Hard to follow logic. Please document what is the execution logic, or use a standard function to simplify the expression

quaat (changes requested):

src/dlite-type.c
jesper-friis: Good point. Done

src/utils/strutils.c
jesper-friis: Added some comments to make it easier to follow.

src/utils/strutils.c
jesper-friis: Good suggestion. When adding explanations it became clear that the code both could be simplified and made more efficient by removing an unnecessary call to realloc().

quaat (approved):

PR #640, opened 2023-09-10
Fixed issue with creating metadata

jesper-friis:
# Description
Avoid calling `dlite.Instance.__init__()` when creating new metadata using the dlite.Metadata constructor, since there is no need for it (and avoids DLite calling abort() because it ends up in an unexpected case...).

Added test that both tests creation of metadata and that it is possible to create two metadata with ref-properties that cyclic refer to each other.

## Type of change
- [x] Bug fix & code cleanup
- [ ] New feature
- [ ] Documentation update
- [x] Test update

## Checklist for the reviewer

### DIFF
--- a/bindings/python/dlite-entity-python.i
+++ b/bindings/python/dlite-entity-python.i
@@ -39,6 +39,9 @@ class Metadata(Instance):
         return Instance.create_metadata(
             uri, dimensions, properties, description)
 
+    def __init__(self, *args, **kwargs):
+        pass  # do nothing, just avoid calling Instance.__init__()
+
     def __repr__(self):
         return f"<Metadata: uri='{self.uri}'>"
 
--- a/bindings/python/tests/test_ref_create.py
+++ b/bindings/python/tests/test_ref_create.py
@@ -1,0 +1,24 @@
+import dlite
+
+
+# Create two entities with ref properties that cyclic refer to each other...
+prop_a = dlite.Property("a", "ref", "http://onto-ns.com/meta/0.1/A")
+B = dlite.Metadata(
+    "http://onto-ns.com/meta/0.1/B", dimensions=[], properties=[prop_a],
+)
+
+prop_v = dlite.Property("v", "int")
+prop_b = dlite.Property("b", "ref", "http://onto-ns.com/meta/0.1/B")
+A = dlite.Metadata(
+    "http://onto-ns.com/meta/0.1/A", dimensions=[], properties=[prop_v, prop_b],
+)
+
+# Create instances
+b = B()
+a = A(properties={"v": 3, "b": b})
+b.a = a
+
+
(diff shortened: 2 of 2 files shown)


bindings/python/tests/test_ref_create.py
```diff
@@ -0,0 +1,24 @@
...
+
+# Create instances
+b = B()
+a = A(properties={"v": 3, "b": b})
+b.a = a
+
+
+assert a.v == 3
+assert a.b == b
+assert b.a == a
```
ajeklund, line 24:
```suggestion
inst_b = B()
inst_a = A(properties={"v": 3, "b": inst_b})
inst_b.a = inst_a


assert inst_a.v == 3
assert inst_a.b == inst_b
assert inst_b.a == inst_a
```

ajeklund: It was a bit complicated to understand the test when several things use the same names `a` and `b`. Here is a suggestion which I think helps a bit.

bindings/python/tests/test_ref_create.py
```diff
@@ -0,0 +1,24 @@
+import dlite
+
+
+# Create two entities with ref properties that cyclic refer to each other...
```
ajeklund, line 4:
```suggestion
# Create two entities (with ref properties) that cyclically refer to each other.
```

bindings/python/dlite-entity-python.i
```diff
@@ -39,6 +39,9 @@ class Metadata(Instance):
         return Instance.create_metadata(
             uri, dimensions, properties, description)
 
+    def __init__(self, *args, **kwargs):
+        pass  # do nothing, just avoid calling Instance.__init__()
```
ajeklund, line 43: Why cannot the constructor of the subclass use the constructor of the superclass at all? To me, removing logic (i.e., functionality) means that the subclass is probably doing less than the superclass, which in turn implies that the class hierarchy might be technically incorrect. What is your thought about this?

ajeklund (changes requested):
Thank you for the PR. I have one question and a couple of small suggestions.

bindings/python/dlite-entity-python.i
jesper-friis: You have a point.

This is a simple pragmatic workaround for that SWIG does not considers subclassing of a SWIG-generated class.

This workaround avoids that the SWIG-internal function dlite.Instance.__init__() calls abort() because it gets a pointer to a dlite.Metadata instead of a dlite.Instance, which it expects.

ajeklund: OK. You might want to consider to describe this SWIG behavior in a comment for this constructor (for future reference).

jesper-friis: Added!

Issue #642, opened 2023-09-12
Document builtin storage plugins


jesper-friis:
Add a page in the users guide about provided storage plugins. 

It should be possible to generates this page based on the documentation of the plugins as shown in `test_storage_plugins.py`.

Ideally the we should also allow extraction of documentation from C. Documentation of options should be possible to automatically extract from the option list.

PR #643, opened 2023-09-12
Iterate over storage plugins

jesper-friis:
# Description
Fixed iteration over storage plugins. One use case is automatic generation of documentation of storage plugins (see issue #642).

Quite many changes, but the most interesting change is [this line](https://github.com/SINTEF/dlite/pull/643/files#diff-afaeddf3751c622eaae4bed224fa4dec3b7db2c6d95d86b0a837a1b3eaeaef17R220).

Additional fixes:
- Got rid of occasional segfaults during shutdown of DLite.
- Updated tests

## Type of change
- [x] Bug fix & code cleanup
- [ ] New feature
- [ ] Documentation update
- [x] Test update

## Checklist for the reviewer


bindings/python/dlite-storage.i
```diff
@@ -49,8 +49,12 @@ Returns next instance or None if exhausted.") next;
   %newobject next;
   struct _DLiteInstance *next(void) {
     char uuid[DLITE_UUID_LENGTH+1];
-    if (dlite_storage_iter_next($self->s, $self->state, uuid) == 0)
-      return dlite_instance_load($self->s, uuid);
+    if (dlite_storage_iter_next($self->s, $self->state, uuid) == 0) {
+      DLiteInstance *inst = dlite_instance_load($self->s, uuid);
+      // Why isn't the refcount already increased?
```
lovfall, line 54: Is this a potential bug somewhere?

bindings/python/tests/test_postgresql1_write.py
```diff
@@ -54,14 +54,14 @@ def ping_server(server="localhost", port=5432, timeout=3):
 ping_server()
 
 # Add metadata to search path
-dlite.storage_path.append(f"{thisdir}/Person.json")
+dlite.storage_path.append(f"{thisdir}/*Person.json")
```
lovfall, line 57: This being a test and all, I would add both files explicitly instead of using a wildcard. If you want to test the wildcard append, create a different test

storages/python/python-storage-plugins/postgresql.py
```diff
@@ -191,7 +191,7 @@ def table_create(self, meta, dims=None):
         ]
         for p in meta["properties"]:
             decl = f'"{p.name}" {to_pgtype(p.type)}'
-            if len(p.dims):
```
lovfall, line 194: This change does not make sense to me. You test for the length of shape, but use dims, as before you tested and used dims

bindings/python/dlite-storage.i
jesper-friis: Good point. Created a new issue about it: #647

bindings/python/tests/test_postgresql1_write.py
jesper-friis: Done. The wildcard and even just specifying the directory are tested elsewhere.

storages/python/python-storage-plugins/postgresql.py
jesper-friis: Sell seen. p.dims  is an backward-compatible alias for p.shape. We should use p.shape

lovfall (approved):
Lokks good now

PR #667, opened 2023-10-11
WIP: Avoid calling free() during atexit

jesper-friis:
# Description
All memory will be freed  at program termination, so there is not reason to explicitly free memory during atexit, except for reducing false positives when tracking down memory leaks with tools like valgrind. 

This PR changes DLite such that we do not free the session and storage iterator if the environment variable `DLITE_ATEXIT_FREE` is not set. 

We also removed the dependency of psycopg2-binary which can lead to segfault.

## Type of change
- [x] Bug fix & code cleanup
- [ ] New feature
- [ ] Documentation update
- [ ] Test update

## Checklist for the reviewer


jesper-friis:
Closing this PR, since the last changes that we want to keep from this PR were applied in PR #736.

PR #676, opened 2023-10-23
Ensure that only metadata is callable

jesper-friis:
# Description
The `__call__()` method only applies to metadata, so moved it from Instance to Metadata.

Fixes the issue with `TypeError: data instances are not callable` in the Musicode demo.

## Type of change
- [x] Bug fix & code cleanup
- [ ] New feature
- [ ] Documentation update
- [ ] Test update

## Checklist for the reviewer

### DIFF
--- a/bindings/python/dlite-entity-python.i
+++ b/bindings/python/dlite-entity-python.i
@@ -50,6 +50,45 @@ class Metadata(Instance):
     def __repr__(self):
         return f"<Metadata: uri='{self.uri}'>"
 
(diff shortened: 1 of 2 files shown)


jesper-friis:
> Moving the **call** method to Metadata makes sense since it only applies to it. However, I can't find to test for it, can you point me to it? In test we should also add something that checks the issue the popped up in the musicode demo (to make sure it does not pop up again).

Added a simple test

francescalb (approved):

PR #681, opened 2023-10-28
Build dlite docker containers for python 3.12

jesper-friis:
# Description
Build dlite docker containers for python 3.12

## Type of change
- [ ] Bug fix & code cleanup
- [x] New feature
- [ ] Documentation update
- [ ] Test update

## Checklist for the reviewer

### DIFF
--- a/.github/docker/Dockerfile-manylinux.template
+++ b/.github/docker/Dockerfile-manylinux.template
@@ -32,8 +32,8 @@ COPY requirements_full.txt requirements_dev.txt /tmp/
 {{ EXTRA_PRE }}
 
 # Enable rpmfusion for additional packages
-RUN yum update -y && \
-  yum localinstall -y --skip-broken \
+RUN yum update -y
+RUN yum localinstall -y --skip-broken \
     https://mirrors.rpmfusion.org/free/el/rpmfusion-free-release-$(rpm --eval %{centos_ver}).noarch.rpm \
     https://mirrors.rpmfusion.org/nonfree/el/rpmfusion-nonfree-release-$(rpm --eval %{centos_ver}).noarch.rpm && \
   yum install -y \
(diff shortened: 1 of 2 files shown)


.github/docker/Dockerfile-manylinux.template
```diff
@@ -42,18 +42,19 @@ RUN yum update -y && \
     swig \
     libcurl-devel \
     libxslt-devel \
-    libxml2-devel && \
-  if [ "{{ TYPE }}{{ ARCH }}" != "2014i686" ]; then yum install -y hdf5-devel; fi && \
-  if [ -f "/etc/yum.repos.d/pgdg-91.repo" ]; then yum install -y postgresql91-devel; fi && \
+    libxml2-devel
+RUN yum install -y rust cargo || true
+RUN if [ "{{ TYPE }}{{ ARCH }}" != "2014i686" ]; then yum install -y hdf5-devel; fi
+RUN if [ -f "/etc/yum.repos.d/pgdg-91.repo" ]; then yum install -y postgresql91-devel; fi
```
francescalb, line 48: I guess these are required specifically for python3.12 only, since they are added now?
If so, I think is is good to write a comment about this for each line.

.github/docker/Dockerfile-manylinux.template
```diff
@@ -42,18 +42,19 @@ RUN yum update -y && \
     swig \
     libcurl-devel \
     libxslt-devel \
-    libxml2-devel && \
-  if [ "{{ TYPE }}{{ ARCH }}" != "2014i686" ]; then yum install -y hdf5-devel; fi && \
-  if [ -f "/etc/yum.repos.d/pgdg-91.repo" ]; then yum install -y postgresql91-devel; fi && \
+    libxml2-devel
+RUN yum install -y rust cargo || true
```
francescalb, line 46:
```suggestion
# Rust and cargo are needed by pydantic2, but may not be available
RUN yum install -y rust cargo || true
```

francescalb (changes requested):
Please update with comments so that the reason for added packages is clear. Otherwise ready to approve.

.github/docker/Dockerfile-manylinux.template
francescalb: If it is not needed I think we should remove it. If the ci.workflow does not fail I think it is safe to remove it.

jesper-friis: done

francescalb (approved):
It would ill be nice to add a comment on why the hdf5 package is needed (refer to the hpd5 plugin), but otherwise very good.

jesper-friis:
> It would ill be nice to add a comment on why the hdf5 package is needed (refer to the hpd5 plugin), but otherwise very good.

Comment added. Waiting for the CI/CD to complete, then I will merge. Thanks

PR #684, opened 2023-10-29
Avoid running TEM demo as a test if dependencies are not fulfilled

jesper-friis:
# Description
Added more dependency checks before running TEM demo as a test. 

Extra requirements needed by the TEM demo have been added to requirements_full.txt.

## Type of change
- [ ] Bug fix & code cleanup
- [ ] New feature
- [ ] Documentation update
- [x] Test update

## Checklist for the reviewer

### DIFF
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -3,6 +3,6 @@ mongomock>=4.1.2,<5
 openpyxl>=3.0.9,<3.2
 
 # Requirements for building wheel
-oldest-supported-numpy>=2023.8.3
+#oldest-supported-numpy>=2023.8.3
 cmake>=1.21.6
 pytest>=7,<8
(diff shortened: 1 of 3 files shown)


requirements_dev.txt
```diff
@@ -3,6 +3,6 @@ mongomock>=4.1.2,<5
 openpyxl>=3.0.9,<3.2
 
 # Requirements for building wheel
-oldest-supported-numpy>=2023.8.3
+#oldest-supported-numpy>=2023.8.3
```
francescalb, line 6: When was this introduced and why do you remove it now just with a comment?

jesper-friis: It results in dependency conflicts with the numpy requirement in requirements.txt. 

I think we can remove it completely. If a normal user wants the development dependencies, there is no need to require the oldest-supported-numpy. oldest-supported-numpy should only be used internally in the docker environment where we build the wheel.

francescalb (approved):
Good

PR #687, opened 2023-10-30
Added test utilities

jesper-friis:
# Description
Replaced pytest.raises() with an own implementation. Another benefit is that it silence expected DLite error messages written to stderr.

Also removed some unused imports in swig interface. OrderedDict is not longer needed after Python 3.7, which is the lowest version of Python supported by DLite.

Alse allowed dlite.errctl() hide and show arguments to be boolean (for convenience).

Also added some extra tests to reproduce a segfault (issue #686). 

## Type of change
- [x] Bug fix & code cleanup
- [ ] New feature
- [ ] Documentation update
- [x] Test update

## Checklist for the reviewer
This checklist should be used as a help for the reviewer.

- [ ] Is the change limited to one issue?
- [ ] Does this PR close the issue?
- [ ] Is the code easy to read and understand?
- [ ] Do all new feature have an accompanying new test?
- [ ] Has the documentation been updated as necessary?

PR #688, opened 2023-10-30
Fixed segfault caused by misspelled shape

### LINKED ISSUE #686: Misspelling of shape in the datamodel may cause DLite to crash
To reproduce the segfault, uncomment the two last lines in the following code segment at end of bindings/python/tests/test_entity.py:

```python
# Test loading invalid json input
with dlite.errctl(hide=True):
    Invalid1 = dlite.get_instance("http://onto-ns.com/meta/0.1/Invalid1")
with raises(dlite.DLiteMissingInstanceError, dlite.DLiteSyntaxError):
    invalid1 = Invalid1([2], properties={"name": "a", "f": [3.14, 2.72]})

# For issue #686. Uncommenting the two last lines will result in segfault
Invalid2 = dlite.get_instance("http://onto-ns.com/meta/0.1/Invalid2")
#with raises(dlite.DLiteMissingInstanceError, dlite.DLiteSyntaxError):
#    invalid2 = Invalid2([2])
```

It seems that the segfault is related to invalid shape name of a 'ref' type property, since Invalid1 produces an expected exception instead of causing a segfault.

jesper-friis:
# Description
#closes #686

I comment on the fix here to avoid clutter the code with too many comments.

**Short motivation for the change in dlite-entity.c**:
-  `memptr` points to the start of the memory segment of the array data
-  if `memptr` is NULL, the array has not been allocated and there is nothing to free (trying to free it will cause segfault)
- the if statement on line 583 ensures that dlite_type_clear() is only called when there is something to free

The reason that a misspelled shape name results in `memptr` being NULL is that `_instance_propdims_eval()` (dlite-entity.c:455) will fail. This will jump directly to line 495 where the refcount of the instance we are trying to create will be decerased. Since this refcount is already zero, `dlite_instance_free()` will be called with a partly instantiated instance.

**NOTE**: This PR builds on top of PR #687. Merge that PR first.

## Type of change
- [x] Bug fix & code cleanup
- [ ] New feature
- [ ] Documentation update
- [ ] Test update

## Checklist for the reviewer

### DIFF
--- a/bindings/python/tests/test_entity.py
+++ b/bindings/python/tests/test_entity.py
@@ -312,3 +312,15 @@ schema.meta.save("basic_metadata_schema.json?mode=w;arrays=false")
 
 mm = dlite.Instance.from_url("json://entity_schema.json")
 assert mm.uri == dlite.ENTITY_SCHEMA
+
+
+# Test loading invalid json input
+with dlite.errctl(hide=True):
+    Invalid1 = dlite.get_instance("http://onto-ns.com/meta/0.1/Invalid1")
+with raises(dlite.DLiteMissingInstanceError, dlite.DLiteSyntaxError):
+    invalid1 = Invalid1([2], properties={"name": "a", "f": [3.14, 2.72]})
+
+# For issue #686
+Invalid2 = dlite.get_instance("http://onto-ns.com/meta/0.1/Invalid2")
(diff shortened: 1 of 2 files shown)


francescalb (approved):
OK

PR #694, opened 2023-11-03
Added uninstall cmake target

jesper-friis:
# Description
Added a uninstall cmake target, making it easy to uninstall dlite when it is installed with cmake. Just do

    make uninstall

or with cmake

    cmake --build . --target uninstall

This is simply a slightly adapted copy of the recipe from the public cmake documentation: https://gitlab.kitware.com/cmake/community/-/wikis/FAQ#can-i-do-make-uninstall-with-cmake

## Type of change
- [ ] Bug fix & code cleanup
- [x] New feature
- [ ] Documentation update
- [ ] Test update

## Checklist for the reviewer

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -829,3 +829,15 @@ if(CROSS_TARGET)
     DESTINATION "lib"
     )
 endif()
+
+
+# Uninstall
+if(NOT TARGET uninstall)
+  configure_file(
+    "${CMAKE_CURRENT_SOURCE_DIR}/cmake/uninstall.cmake.in"
+    "${CMAKE_CURRENT_BINARY_DIR}/uninstall.cmake"
+    IMMEDIATE @ONLY)
+
+  add_custom_target(uninstall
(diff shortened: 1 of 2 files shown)


francescalb (changes requested):
Does not work for me, I can still import dlite after having run either of

`cmake --build . --target uninstall` or `make uninstall`

I am working in an environment and everything is still in the environement. Deactivating and reactivating does not change anything either.

jesper-friis:
> Does not work for me, I can still import dlite after having run either of
> 
> `cmake --build . --target uninstall` or `make uninstall`
> 
> I am working in an environment and everything is still in the environement. Deactivating and reactivating does not change anything either.

You are right, it didn't removed directories. It seems that cmake doesn't record installed directories. 

To also uninstall directories, I added some extra code lines that after all installed files are removed check the parent directory of the installed files. If the directory is empty, it is removed.

This should have the expected behaviour in almost all cases, but could potential remove an empty directory in the installation path that existed before dlite was installed.

francescalb (approved):
Seems to work nicely.

PR #695, opened 2023-11-04
Updated gitignore files

jesper-friis:
# Description
Cleaned up .gitignore files.

## Type of change
- [x] Bug fix & code cleanup
- [ ] New feature
- [ ] Documentation update
- [ ] Test update

## Checklist for the reviewer

### DIFF
--- a/python/.gitignore
+++ b/python/.gitignore
@@ -1,44 +1,6 @@
 *~
 
-external
-build*/
-install*/
 dist/
 wheelhouse/
-local/
-jansson
-hdf5-1.10.*
-.vscode/
-.gdb_history
-dlite.egg-info
-
-# Ignore personal modified bootstrap files
-bootstrap-win-*.sh
-
-# For Visual Studio
-.vs
-CMakeSettings.json
-
-# For qtcreator
-CMakeLists.txt.user
-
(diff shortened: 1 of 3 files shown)


python/.gitignore
```diff
@@ -1,44 +1,6 @@
...
-
-# Generated build docker files
-Dockerfile-*linux*[^.template]
-
-# Units (pint) files
-units.cpython-*.pyc
-bindings/python/triplestore/__pycache__
-
-# Documentation
-pydoc/autoapi/
```
francescalb, line 41: Why remove this?

francescalb:
Is the reason for this clean-up is to move gitignore to the parent directory? 

Also, there are quite a few thigns that are not only moved but removed, why?

python/.gitignore
jesper-friis: They are included in the top-level gitignore file.

francescalb (approved):

jesper-friis:
> Is the reason for this clean-up is to move gitignore to the parent directory?
> 
> Also, there are quite a few thigns that are not only moved but removed, why?

Most of what is removed was an unnecessary duplication of the toplevel .gitignore file. There were also some too wide entries that matched too much.

PR #700, opened 2023-11-04
Do not free the librdf world if called from atexit handler.

jesper-friis:
# Description
Do not free the librdf world if called from atexit handler.

This may avoid a segfault due to inconsistent state during termination. It is not possible to add a test for this because it depends on the order that the operating system is freeing resources.

## Type of change
- [x] Bug fix & code cleanup
- [ ] New feature
- [ ] Documentation update
- [ ] Test update

## Checklist for the reviewer


francescalb (approved):
Approved, but please write a little bit more info in the merge commit so that this line changed can be understood at a later date.

PR #708, opened 2023-11-13
Created a human readable error message for missing units.

jesper-friis:
# Description
Created a human readable error message for missing units.

Also added a ad-hock code for adding units. This is a dirty workaround that should be done properly.

## Type of change
- [x] Bug fix & code cleanup
- [ ] New feature
- [ ] Documentation update
- [ ] Test update

## Checklist for the reviewer


bindings/python/mappings.py
```diff
@@ -72,14 +73,35 @@ def instance_routes(
...
 
+    # FIXME: Bad workaround for adding missing units to the registry.
+    # Too be fixed ASAP.
+    ureg = quantity(1.0, "m").units._REGISTRY
+    unit_definitions = [
+        ("atom_fraction", "atom_fraction = 1"),
+    ]
+    for uname, udef in unit_definitions:
+        if uname not in ureg:
+            ureg.define(udef)
```
francescalb, line 84: I do not understand why we hardcode the addition of a missing unit here. Should we not rather allow for the addition of a unit by the user if needed?

bindings/python/mappings.py
```diff
@@ -72,14 +73,35 @@ def instance_routes(
...
         for key, value in inst.properties.items():
             if props[key].unit:
+
+                try:
+                    q = quantity(1.0, props[key].unit)
+                except TypeError:
+                    raise UnknownUnitError(
+                        f"unknown unit '{props[key].unit}' in datamodel: "
+                        f"{inst.meta.uri}"
+                    )
```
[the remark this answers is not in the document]
jesper-friis: Yes. Pint raises consistently TypeError from deep down in its code if the unit is unknown.

A better solution would be that the functions in the mappings module takes a pint.UnitRegistry (or maybe better a class wrapped around it) instead of pint.Quantity class as input. Then we could replace the above try-except with

    if props[key].unit not in ureg:
        raise UnknownUnitError(...)

bindings/python/mappings.py
jesper-friis: Definitely. This just a quick hacky workaround to add atom_fraction to the pint unit registry such that the ss3 test works.

francescalb: I really don't think we should putin quick and dirty fixes like this into the main DLite. How much work is it to find a proper solution?

jesper-friis: Ok, removed the hack. The price for that is that SS3 will not work...

A proper solution will probably take 6h to implement. It will mainly involve tripper (issue https://github.com/EMMC-ASBL/tripper/issues/154) with some fixes to DLite as well.

bindings/python/mappings.py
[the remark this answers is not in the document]
francescalb: So in which test it this being tested?

francescalb (approved):
Please make an issue in SS3 for OpenModel on how we agreed to solve this (or just do it if that is quicker).

PR #712, opened 2023-11-22
WIP: Fix segfaults

jesper-friis:
# Description
This PR adds a set of fixes to avoid segfaults:
  - avoid crash if session_add_state() is called while freeing a session (this can happen if the free function calls session_add_state())
  - interpret when Python is shutting down like we are in an atexit handler
  - replaced the (mis)use of ATEXIT_MARKER_ID to set/check whether we are in an atexit handler with two clean function calls: - dlite_globals_set_atexit() - dlite_globals_in_atexit()
  - added a local state in dlite-misc.c
  - improved error codes in dlite-storage.c

This PR applies some of the fixes in PR #667, while trying to avoid the problems. However, not all the fixes in #667 are addressed in this PR, so it is still too early to close #667.

## Type of change
- [x] Bug fix & code cleanup
- [ ] New feature
- [ ] Documentation update
- [ ] Test update

## Checklist for the reviewer

### DIFF
--- a/bindings/python/dlite-misc-python.i
+++ b/bindings/python/dlite-misc-python.i
@@ -4,6 +4,11 @@
 
 %pythoncode %{
 
+import atexit
+
+atexit.register(_mark_python_atexit)
+
+
 class errctl():
     """Context manager for temporary disabling specific DLite error
     messages or redirecting them.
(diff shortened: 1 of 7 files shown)


jesper-friis:
All changed addressed in PR #716 and #719.

PR #715, opened 2023-11-25
Updated setting default CMAKE_INSTALL_PREFIX such that it is only set once

jesper-friis:
# Description
Updated setting default CMAKE_INSTALL_PREFIX such that it is only set once.

Hopefully this will work on all systems.

## Type of change
- [x] Bug fix & code cleanup
- [ ] New feature
- [ ] Documentation update
- [ ] Test update

## Checklist for the reviewer

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -245,12 +245,6 @@ if(WITH_PYTHON)
   if(DEFINED ENV{VIRTUAL_ENV})
     message(STATUS "Detected virtual environment $ENV{VIRTUAL_ENV}")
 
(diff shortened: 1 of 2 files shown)


jesper-friis:
> Works on my setup now. For me it is OK, but I wonder if maybe our test system is not sufficient to check this longterm. Anyhow, it works, so I approve :)

Great that it works. I think that this PR is a longterm solution for selecting a correct default value for CMAKE_INSTALL_PREFIX. I think the best way to test the build system is to just use it and be responsive if anyone has issues. The CI build tests will only test for the build github build environment that always starts out clean and without previous installations of DLite.

PR #716, opened 2023-11-26
Added refcounting to storages

jesper-friis:
# Description
The reason that test_python_bindings sometimes segfault seems to be that the storage is closed while still in use in a storage iterator. To avoid that, we added reference counting to DLite storages, such that they are first freed when there are no more references to it.

Furthermore this PR marks when entering an Python atexit handler.

While Python is shutting down, it may leave DLite in an inconsistent state. By notify the DLite that we are in an atexit handler, we can skip freeing memory that may cause segmentation fault, without having to bother about memory leaks (since the system is shutting down and all memory will be freed anyway).

## Type of change
- [x] Bug fix & code cleanup
- [ ] New feature
- [ ] Documentation update
- [ ] Test update

## Checklist for the reviewer


francescalb (approved):
Approived under the condition that a through merge commit message is written that describes what is happening here and why. This is because we are not adding a test specifically, but we have had tests that fail randomly and hence it will be difficult to understand the reasoning for this PR at a later stage.

PR #717, opened 2023-11-26
Notify the C code when entering a Python atexit handler

jesper-friis:
# Description
Mark when entering an Python atexit handler.

While Python is shutting down, it may leave DLite in an inconsistent state. By notify the DLite that we are in an atexit handler, we can skip freeing memory that may cause segmentation fault, without having to bother about memory leaks (since the system is shutting down and all memory will be freed anyway).

## Type of change
- [x] Bug fix & code cleanup
- [ ] New feature
- [ ] Documentation update
- [ ] Test update

## Checklist for the reviewer

### DIFF
--- a/bindings/python/dlite-misc-python.i
+++ b/bindings/python/dlite-misc-python.i
@@ -4,6 +4,11 @@
 
 %pythoncode %{
 
+import atexit
+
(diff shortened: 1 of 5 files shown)


jesper-friis:
Better fix included in PR #716.

Issue #609, opened 2023-08-20
Move test input/output to separate directories


jesper-friis:
Currently running the tests pollutes the current directory with a lot of test output files. To avoid this, write test output to a `output` sub-directory of the test directory. 

It would also be easier to spot test input if they are moved to a `input` sub-directory of the test directory. 

Test directories that should be fixed:
- [ ] bindings/python/tests/
- [ ] bindings/fortran/tests
- [ ] src/tests
- [ ] src/utils/tests
- [ ] storages/json/tests
- [ ] storages/rdf/tests
- [ ] tools/tests

jesper-friis:
PR #720 addresses this for two tests in binding/python/tests/

PR #720, opened 2023-11-26
Applied general fixes from PR 667

jesper-friis:
# Description
Applied general improvements from PR 667.

The purpose of this PR is to ensure that the fixes in PR 667 are not lost. This includes
- save test output in a separate directory (for some tests, see issue #609)
- do not run test_property_mappings if rdflib is missing
- documented environment variable DLITE_ATEXIT_FREE
- fixed memory leak in test_bson_storage

## Type of change
- [x] Bug fix & code cleanup
- [ ] New feature
- [ ] Documentation update
- [ ] Test update

## Checklist for the reviewer


bindings/python/tests/test_pydantic.py
```diff
@@ -60,7 +60,7 @@ class TransformationStatus(BaseModel):
 assert inst.startTime == int(now - 3000)
 utc = timezone(timedelta(hours=0))
 dt = datetime.fromtimestamp(now - 600).astimezone(utc)
-assert inst.finishTime == str(dt)
+assert inst.finishTime.split("+")[0] == str(dt).split("+")[0]
```
francescalb, line 63: Please comment why

cmake/valgrind-python.supp
francescalb, line 1: Describe the purpose of this.

storages/json/dlite-json-storage.c
francescalb, line 1: comment on in commit

bindings/python/tests/test_pydantic.py
jesper-friis: done

cmake/valgrind-python.supp
jesper-friis: done

storages/json/dlite-json-storage.c
jesper-friis: Avoid memory leak by calling `fu_endmatch()` when we are done with the iterator. 

To not polute the code with unnecessary comments, I prefer not to add this comment to the code since it is already documented in `fu_startmatch()` that the returned iterator must be cleaned up with `fu_endmatch()`.

src/dlite-storage-plugins.c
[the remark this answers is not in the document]
jesper-friis: To avoid memory leak by calling fu_endmatch() when we are done with the iterator.

To not polute the code with unnecessary comments, I prefer not to add this comment to the code since it is already documented in fu_startmatch() that the returned iterator must be cleaned up with fu_endmatch().

francescalb (approved):
OK. We disagree a bit on where comments should be. I guess that is because I am not enough expert in C to understand the code by itself.

Issue #743, opened 2023-12-08
Add unit support in property access and assignment


jesper-friis:
Issue suggested by @terence-coudert.

Suggested interface for assignment:
```
inst.stress = "3 MPa"  # this will convert the str to a Quantity and store the value in the right unit
inst.stress = 3   # this is already implemented
inst.stress = ([3.0, 4.0, 5.0], "MPa")  # this convert the tuple into a Quantity
inst.stress = Quantity(70, "GPa")
```

Suggested interface for access:
```
inst.stress.to("MPa") -> 70e3
```

PR #744, opened 2023-12-09
Unit support when accessing and assigning properties

jesper-friis:
# Description
Addresses issue #743

Added support for specifying when accessing and assigning properties. The behaviour is not exactly as suggested in issue #743.

Currently this PR only adds support for assignment with a pint Quantity:
```
inst.stress = 3   # this is already implemented
inst.stress = Quantity(70, "GPa")  # added in this PR
```
Support for the other cases could be added to the new `set_property()` in bindings/python/dlite-entity-python.i.

For access issue #743 suggested the syntax `inst.stress.to("MPa")`. However, `inst.stress` is the value of the property (a scalar (e.g. float) or a numpy array). Unless that we want to create our own subclasses of basic scalars and numpy.array, it is not possible to add a `to()` method. Instead PR adds the python property `dlite.Instance.q` for accessing the properties and pint.Quantity objects. For example, this will work:
```
inst.q.stress  -->  Quantity(70, "GPa")
inst.q.stress.to("MPa")  --> Quantity(70e3, "MPa")
inst.q.stress.to("MPa").m  -->  70e3
```

Unrelated to the issue, also made some exceptions more specific. 

## Type of change
- [ ] Bug fix & code cleanup
- [x] New feature
- [ ] Documentation update
- [ ] Test update

## Checklist for the reviewer


terence-coudert:
I did an implementation to full fill the following statements:
```python
inst.q.stress  -->  Quantity(70, "GPa")
inst.q.stress.to("MPa")  --> Quantity(70e3, "MPa")
inst.q.stress.to("MPa").m  -->  70e3
inst.q.stress = "70 GPa"
inst.q.stress = (70.0,  "GPa")
inst.q.stress = inst.q.quantity(500, "kN") / inst.q.quantity(10, "mm^2")
inst.set_quantity("stress", 70.0, "GPa")
inst.get_quantity("stress") --> Quantity(70, "GPa")
```
I made a series of unit tests (see bindings/python/tests/test_quantity.py)

The following statement is not yet implemented:
```python
inst.stress = Quantity(70, "GPa")
```
The proposed implementation here does not use your proposal (the class _QuantityProperty).

terence-coudert:
Any custom UnitRegistry can be used with the proposed implementation. The user needs to set the UnitRegistry by the following:
```python
import pint
pint.set_application_registry(my_unit_registry)
# the current unit registry is also accessible by
inst.q.unit_registry
inst.q.temperature = "-20 degC"
```

terence-coudert:
Maybe we should not implement the following statement, I think it's not really consistent if we can assign a Quantity to a property but we cannot get a Quantity from a property. I suggest to use always the Python property "q" when working with quantity. I have added a section in the documentation see [tutorial.md](https://github.com/SINTEF/dlite/blob/743-property-units/doc/getting_started/tutorial.md#working-with-physical-quantities)
```python
inst.stress = Quantity(70, "GPa")
# maybe the user will expect a Quantity when doing
s = inst.stress
```

bindings/python/quantity.py
```diff
@@ -0,0 +1,175 @@
...
+    'float32', 'float64', 'float80', 'float96', 'float128'
+]
+quantity_helper = None
+
+
+class QuantityHelper:
+
+    def __init__(self):
+        self.__dict__['_instance'] = None
+        self.__dict__['_registry'] = None
```
jesper-friis, line 22: Why can't you just write it like 

    self._instance = None
    self._registry = None

bindings/python/quantity.py
jesper-friis, line 1: Please add a short module docstring describing the purpose of this module.

README.md
```diff
@@ -128,6 +128,18 @@ In this case we didn't initialised the age
 >>> homes.age = 34  # Assign the age
 ```
 
+If you have [Pint] installed, you can also specifying or accessing the age
```
jesper-friis, line 131:
```suggestion
If you have [Pint] installed, you can also specify or access the age
```

bindings/python/quantity.py
terence-coudert: because I'm overwriting the method __setattr__, and doing ```self._instance = None``` will raise a RecursionError

PR #755, opened 2023-12-26
Add datatype to collection

### LINKED ISSUE #741: Add datatype to collection
Collections are currently only a list of s,p,o triples, represented as three strings. However, in order to represent RDF graphs as relations in a collection, we also have to store the datatype of literals. 

This issue relates closely to issue #174

jesper-friis:
# Description
Add datatype to collections. This is needed for representing RDF literals.

Also added more tests and fixed bugs.

Closes #741

## Type of change
- [x] Bug fix & code cleanup
- [x] New feature
- [ ] Documentation update
- [x] Test update

## Checklist for the reviewer


bindings/python/dlite-collection-python.i
```diff
@@ -49,9 +57,17 @@ class Collection(Instance):
...
     Use the get_relations() method (or the convenience methods
-    get_subjects(), get_predicates() and get_predicates()) to iterate
-    over relations.  The number of relations is available via the `nrelations`
-    property.
+    get_subjects(), get_predicates() and get_objects()) to iterate
+    over relations.  The number of relations is available via the
+    `nrelations` property.
+
+    Relations are (s, p, o, d=None)-triples with an optional fourth field
+    `d`, specifying the datatype of the object.  It may have the following
```
francescalb, line 65: It is what? It is a bit unclear what you are referring to.  From the description nelow it seems you are referring to 'object', but it is a bit unclear.

bindings/python/dlite-collection-python.i
```diff
@@ -191,21 +207,22 @@ class Collection(Instance):
...
+            raise _dlite.DLiteError(
+                f'Error adding relation ({s}, {p}, {o}, {d})')
 
-    def remove_relations(self, s=None, p=None, o=None):
+    def remove_relations(self, s=None, p=None, o=None, d=None):
         """Remove all relations matching `s`, `p` and `o`."""
-        if _collection_remove_relations(self._coll, s, p, o) < 0:
+        if _collection_remove_relations(self._coll, s, p, o, d) < 0:
             raise _dlite.DLiteError(
                 f'Error removing relations matching ({s}, {p}, {o})')
```
francescalb, line 220:
```suggestion
                f'Error removing relations matching ({s}, {p}, {o}, {d})')
```

bindings/python/dlite-collection-python.i
```diff
@@ -11,24 +11,32 @@
...
             raise StopIteration()
 
-        if self.rettype == 'T':  # return (s,p,o) tuple
+        if self.rettype == 'I':  # return instance
+            return v
+        if self.rettype == 'R':  # return relation object
+            return v.copy()
+        if self.rettype == 'T':  # return (s,p,o,d) tuple
+            return (v.s, v.p, v.o, v.d)
+        if self.rettype == 't':  # return (s,p,o) tuple
```
francescalb, line 28: When introducing a new type of variable, isn't it better that the 'old' one stays the same as before?

bindings/python/dlite-collection-python.i
```diff
@@ -295,26 +312,54 @@ class Collection(Instance):
...
+
+    def value(self, s=None, p=None, o=None, d=None, default=None, any=False):
+        """Return pointer to the value for a pair of two criteria.
 
+        Useful if one knows that there may only be one value.
+
+        Parameters:
+            s, p, o: Criteria to match. Two of these must be non-NULL.
+            d: If not NULL, the required datatype of literal objects.
+            default: Value to return if no matches are found.
```
francescalb, line 355: A bit confusing to start referring to NULL, as this is in python. I think it is beter to use None and if it is importatnt to explain the connection to C, do that.

bindings/python/dlite-collection.i
```diff
@@ -26,13 +41,19 @@ int dlite_collection_save(struct _DLiteCollection *coll,
...
+                                      const char *o, const char *d=NULL);
 const struct _Triple *
   dlite_collection_find_first(const struct _DLiteCollection *coll,
-                              const char *s, const char *p, const char *o);
+                              const char *s, const char *p, const char *o,
+                              const char *d=NULL);
+//const char *dlite_collection_value(struct _DLiteCollection *coll,
+//                                   const char *s=NULL, const char *p=NULL,
+//                                   const char *o=NULL, const char *d=NULL,
+//                                   const char *fallback=NULL, int any=0);
```
francescalb, line 56: Can be deleted

bindings/python/dlite-collection.i
```diff
@@ -86,6 +107,30 @@ struct _DLiteCollection {
...
+  //      The value of the `s`, `p` or `o` that is None.
+  //  ") value;
+  //const char *value(struct _DLiteCollection *coll,
+  //                  const char *s=NULL, const char *p=NULL,
+  //                  const char *o=NULL, const char *d=NULL,
+  //                  const char *fallback=NULL, int any=0) {
+  //  const char *v = triplestore_value(coll->rstore, s, p, o, d, fallback, any);
+  //  return (v) ? strdup(v) : NULL;
+  //}
+
```
francescalb, line 133: I think we should not keep commented out function in the code. 

One questino, though, do I red it correctly in that value search is moved from the core C code to the bindings in Python?

bindings/python/dlite-collection.i
```diff
@@ -97,8 +142,8 @@ struct _DLiteCollection {
   struct _CollectionIter {
     DLiteCollection *coll;
     DLiteCollectionState state;
-    char *s, *p, *o;  /* search pattern */
-    char rettype;     /* return type: I=Instance, R=Relation, T=tuple */
+    char *s, *p, *o, *d;  /* search pattern */
+    char rettype;         /* return type: I,R,T,t,s,p,o,d */
```
francescalb, line 146: Whye do you not have I=Instance, R=Relation, T=tuple, t=tuple.... etc here as before? (much easier for me to read and know what we are considering)

bindings/python/dlite-python.i
```diff
@@ -931,33 +931,37 @@ int dlite_swig_set_scalar(void *ptr, DLiteType type, size_t size, obj_t *obj)
...
           Py_DECREF(lst);
-          FAIL1("relations must be 3 or 4 strings, got %ld", (long)n);
+          FAIL1("relations must be 3 to 5 strings, got %ld", (long)n);
         }
         for (i=0; i<n; i++) {
           PyObject *item = PySequence_Fast_GET_ITEM(lst, i);
           if (!PyUnicode_Check(item)) {
             Py_DECREF(lst);
-            FAIL("relation subject, predicate and object must be strings");
+            FAIL("relation (s,p,o[,d[,id]]) items must be strings");
```
francescalb, line 955: I liked that the were written out in the previous error message. How about using subject(s), predicate(p)...

bindings/python/dlite-python.i
```diff
@@ -969,14 +973,14 @@ int dlite_swig_set_scalar(void *ptr, DLiteType type, size_t size, obj_t *obj)
         }
 
         /* Assign id if not already provided */
-        if (n < 4)
-          s[3] = triple_get_id(NULL, s[0], s[1], s[2]);
+        if (n < 5)
+          s[4] = triple_get_id(NULL, s[0], s[1], s[2], s[3]);
```
francescalb, line 977: isn't n<5 also if datatype is not defined, while id is?

bindings/python/tests/test_collection.py
```diff
@@ -159,6 +161,24 @@
...
+    coll.value(s="dog", p="is-a", o="animal")
+
+with raises(dlite.DLiteTypeError):
+    coll.value()
+
+with raises(dlite.DLiteSearchError):
+    coll.value(s="dog", p="x")
+
+with raises(dlite.DLiteSearchError):
+    coll.value(p="_is-a", o="Instance")
```
francescalb, line 179: What happens if we set d? Do we get a nice error? I think that is something that people could think of doing since we can set s, p, o.

bindings/python/tests/test_relation.py
```diff
@@ -0,0 +1,4 @@
+import dlite
+
+rel1 = dlite.Relation("s1", "p1", "o1")
+rel2 = dlite.Relation("s2", "p2", "o2", "d2")
```
francescalb, line 4: What are we testing here? Looks like we tests that setting a relation does not throw an error, but do we really test that it is there? Also, I would like tests on doing it wrongly as well, to check that the desired Errors are actually raised.

bindings/python/tests/test_storage.py
```diff
@@ -34,6 +34,11 @@
 # Create an instance
 inst = myentity(dimensions=[2, 3], id="my-data")
 inst["a-bool-array"] = True, False
+inst["a-relation"] = "subj", "pred", "obj"
+inst["a-relation-array"] = [
+    ("s1", "p1", "o1"),
+    ("s2", "p2", "o2"),
+]
```
francescalb, line 41: Same as above, can you add a test that checks inst?

src/dlite-collection.c
```diff
@@ -441,14 +460,15 @@ const DLiteRelation *dlite_collection_find(const DLiteCollection *coll,
...
-  r = dlite_collection_find(coll, &state, s, p, o);
-  dlite_collection_deinit_state(&state);
-  return r;
+  return dlite_collection_find(coll, NULL, s, p, o, d);
+  //DLiteCollectionState state;
+  //const DLiteRelation *r;
+  //dlite_collection_init_state(coll, &state);
+  //r = dlite_collection_find(coll, &state, s, p, o, d);
+  //dlite_collection_deinit_state(&state);
+  //return r;
```
francescalb, line 471: Why keep these comments?

src/dlite-collection.c
```diff
@@ -645,7 +669,31 @@ int dlite_collection_count(DLiteCollection *coll)
...
+
+  Parameters:
+      s, p, o: Criteria to match. Two of these must be non-NULL.
+      d: If not NULL, the required datatype of literal objects.
+      fallback: Value to return if no matches are found.
+      any: If non-zero, return first matching value.
+
+  Returns a pointer to the value of the `s`, `p` or `o` that is NULL.
+  On error NULL is returned.
+ */
```
francescalb, line 693: I think this is a nice description. However, I have read several variations on the same theme further up. Does this mean that we are documenting the same thing several places? This is a bit confusing as well as increases the risk of updating documentation of the same thing wrongly (not in all pleaces) if needed later.

src/dlite-collection.h
```diff
@@ -309,4 +316,24 @@ DLiteInstance *dlite_collection_next_new(DLiteCollection *coll,
...
+
+  Parameters:
+      s, p, o: Criteria to match. Two of these must be non-NULL.
+      d: If not NULL, the required datatype of literal objects.
+      fallback: Value to return if no matches are found.
+      any: If non-zero, return first matching value.
+
+  Returns a pointer to the value of the `s`, `p` or `o` that is NULL.
+  On error NULL is returned.
+ */
```
francescalb, line 334: Same comment as above.

src/tests/test_collection.c
```diff
@@ -26,24 +27,28 @@ MU_TEST(test_collection_add_relation)
...
+  mu_check(!dlite_collection_add_relation(coll, "terrier", "is_a", "dog",
+                                          NULL));
   mu_assert_int_eq(3, dlite_instance_get_dimension_size((DLiteInstance *)coll,
 							"nrelations"));
   mu_assert_int_eq(3, coll->nrelations);
 
   /* dublicate... */
-  mu_check(!dlite_collection_add_relation(coll, "terrier", "is_a", "dog"));
+  mu_check(!dlite_collection_add_relation(coll, "terrier", "is_a", "dog",
+                                          NULL));
```
francescalb, line 42: Plase also test setting the datatype.

src/tests/test_collection.c
```diff
@@ -165,29 +238,30 @@ MU_TEST(test_collection_load)
...
+  //dlite_json_print((DLiteInstance *)coll2);
+  //printf("----------------------\n");
+
   const DLiteInstance *inst = dlite_collection_get(coll2, "inst");
-  printf("\n--- inst: %p ---\n", (void *)inst);
-  dlite_json_print((DLiteInstance *)inst);
-  printf("----------------------\n");
+  //printf("\n--- inst: %p ---\n", (void *)inst);
+  //dlite_json_print((DLiteInstance *)inst);
+  //printf("----------------------\n");
```
francescalb, line 265: // can be removed altogether?

src/triplestore-builtin.c
```diff
@@ -382,36 +379,38 @@ const Triple *triplestore_poll(TripleState *state)
...
-  it should be provided a `state` initialised with triplestore_init_state().
+  Return next triple matching s-p-o triple, where `d` is the datatype
+  of the object. Any of `s`, `p`, `o` or `d` may be NULL, in which case
+  they will match anything.
+
+  If `d` starts with '@', it will match language-tagged plain text
+  literal objects whos XML language abbreviation matches the string
+  following the '@'-sign.
+
+  if `d` is NUL (empty string), it will match non-literal objects.
```
francescalb, line 390:
```suggestion
  if `d` is NULL (empty string), it will match non-literal objects.
```

src/triplestore.c
```diff
@@ -6,4 +6,95 @@
...
+
+  Parameters:
+      s, p, o: Criteria to match. Two of these must be non-NULL.
+      d: If not NULL, the required datatype of literal objects.
+      fallback: Value to return if no matches are found.
+      any: If non-zero, return first matching value.
+
+  Returns a pointer to the value of the `s`, `p` or `o` that is NULL.
+  On error NULL is returned.
+ */
```
francescalb, line 62: Again, the same documentation in a new place.

francescalb (changes requested):
I have quite a few comments and questions. Especially I am worried about the documentation being written in so many places. Does this create a risk of wrong updates later?

Also, the tests fail. If this is because an update in tripper is required for it to pass it should not be included in the test.suite for a new release just yet I think. Maybe it is possible to write that the tests should run if tripper is above a certain version (i.e. higher than the current latest).

.github/workflows/ci_tests.yml
```diff
@@ -25,6 +25,7 @@ jobs:
           python3-dev \
           python3-numpy \
           python3-yaml
+          #rdflib0-dev
```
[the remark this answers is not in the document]
jesper-friis: Ok. La inn en kommentar om hvorfor vi ikke inkluderer rdflib0-dev i stedet.

bindings/python/dlite-collection-python.i
jesper-friis: I principle I fully agree. 

But it wasn't long ago that the rettype variable was introduced. It is also quite hidden and was not documented before this PR, so I don't expect anyone using it or even recognised it. Therefore I it is ok to break backward compatibility to the benefit of more logical naming in the future (I associate capital with more).

Another more important question is whether 'T' (s,p,o,d) should be the default for get_relations() rather than 't' (s,p,o). Here I chose 't' for backward compatibility reasons.  But it makes it easy to miss the datatype. I even did it myself when updating tripper. 

What do you think. Is it worth to break compatibility for a probably more safe interface in the future? Maybe a better solution would be to add a new method in the Python interface that also returns the datatype. What would be a good name for such a method?

francescalb: I think T should be default. 
But not sure if we should go for breaking change suddenyl from one release to the next or if we should add warning in some way,

jesper-friis: Agree. Lets do that in a separate PR and check with the team before doing it. I will add an issue.

bindings/python/dlite-collection-python.i
jesper-friis: Updated documentation

bindings/python/dlite-collection-python.i
jesper-friis: Well spotted. I changed it to Python nomenclouture.

bindings/python/dlite-collection.i
jesper-friis: done

bindings/python/dlite-collection.i
jesper-friis: Removed the commented out lines. 

No, value() is still only implemented in C. What I commented out here was my first attempt to interface it directly as a method of the Collection class. I gave that up and added it instead as Python code in dlite-collection-python.i using the low-level swig-generated function in dlite._dlite._collection_value().

bindings/python/dlite-collection.i
jesper-friis: It guess that I was lazy. I updated the comment now.

bindings/python/dlite-python.i
jesper-friis: I think this should work. The datatype has to be provided (possible as None) if id is provided, since a sequence with 5 elements must have a 4th element.

bindings/python/tests/test_collection.py
jesper-friis: Added two additional tests for it. Seems to work :-)

```python
assert coll.value(s="inst1", p="_has-uuid", d="xsd:anyURI") == (
    "de1eabc0-8f8e-4a64-9cd6-74ea4b2e3c54"
)
with raises(dlite.DLiteSearchError):
    coll.value(s="inst1", p="_has-uuid", d="xsd:int")
```

bindings/python/tests/test_relation.py
jesper-friis: Well spotted. Seems that I started on this test script, but forgot to finalise it. Added some more tests now.

bindings/python/tests/test_storage.py
jesper-friis: Ok, added some additional assertments. Had to add a `__eq__()` method to Relation for the test to work.

src/dlite-collection.c
jesper-friis: Yes, we are duplicating documentation. In C we have the same docstring both in the source (.c) and header (.h) file. Function expose in Python will also have a Python documentation. 

It is possible to reduce this duplication. Currently we generate the C reference documentation from the header files, so we need documentation there. At the same time, it is very useful to have the function documentation in the source files, so we also need it there.

A possibility could be to We could remove all function documentation from the header files and update the documentation in the source files such that it can be parsed by doxygen. That would increase maintainability and could be added as an issue.

It would in principle also be possible to extract function documentation from the doxygen-generated xml file and auto-document Python functions from it. But firstly, it would increase the complexity and make dlite even more difficult understand and secondly we want to rephrase the documentation for Python as you commented about above. So this is not something I would suggest.

src/dlite-collection.h
jesper-friis: Yes, this is the header file duplication mentioned in my response above.

src/tests/test_collection.c
jesper-friis: Datatype is already tested in the test_collection_find() function.

src/tests/test_collection.c
jesper-friis: Yes, they could. But loading is rather tricky involving loading a shared library via the plugin system and json-parsing, so it is not unlikely that we have to do some debugging in the future. Then these commented-out print statements comes in handy. I would therefore prefer to keep them.

src/triplestore-builtin.c
jesper-friis: This is some C jargon. NUL means an empty string (a byte of value zero) while NULL means a pointer pointing to the invalid memory address zero (eight zero bytes on a 64-bit machine).

src/triplestore.c
jesper-friis: Yes, we duplicate the documentation of the new `xxx_value()` function a lot. First we document it is triplestore.{c,h} where it is implemented. Then we document it in dlite-collection.{c,h} where it is exposed. Finally in the Python interface. 

I think that it is good to document all functions, even though it results in some duplication. We could strip the documentation from the header files, but that would require to update the doxygen configuration and update the corresponding documentation in the source files. 

Open for other opinions.

jesper-friis:
> I have quite a few comments and questions. Especially I am worried about the documentation being written in so many places. Does this create a risk of wrong updates later?
> 
> Also, the tests fail. If this is because an update in tripper is required for it to pass it should not be included in the test.suite for a new release just yet I think. Maybe it is possible to write that the tests should run if tripper is above a certain version (i.e. higher than the current latest).

It is possible to reduce duplication of documentation, but as mentioned above, it requires to:
- convert documentation in C source files to doxygen comments
- remove documentation from C headers
- update the documentation build system to parse the C sources instead of headers

That should be a separate issue. I am not sure whether it is really is something that should be prioritised.

francescalb (approved):
Nice. Just add testing of including datatype when adding a relation before merging.

PR #792, opened 2024-02-20
Added OTE case to mapping example

jesper-friis:
# Description
Added a new case to `examples/mappings/` showing how to use mapping functions together with OTEAPI. 

## Type of change
- [ ] Bug fix & code cleanup
- [ ] New feature
- [x] Documentation update
- [x] Test update

## Checklist for the reviewer


examples/mappings/requirements.txt
```diff
@@ -0,0 +1,6 @@
+oteapi-core>=0.6.1,<1
+#oteapi-dlite>0.1.6,<1  # update once after migrating to pydantic v2
+oteapi-dlite @ git+https://github.com/EMMC-ASBL/oteapi-dlite.git@cwa/close-187-migrate-to-pydantic-v2
+DLite-Python>=0.5.1,<1
```
francescalb, line 4: This is problematic to include in tests with a different set of requirements. We end up having issues with the dlite-installation when we both install from source and pypi.

examples/mappings/README.md
```diff
@@ -1,15 +1,40 @@
...
-2. Calculating the potential energy and the forces on all atoms using the EMT calculator and storing them in an instance of a **CalcResults** data model.
-3. Mapping properties of the data models and input/output of mapping functions to ontological concepts.
-4. Invoking `dlite.mappings.instantiate()` to instantiate a **Molecule** data model from the mappings.
+
+1. Loading a given atom structure from a data source into an instance
+   of a **ASEAtoms** data model.
+
+2. Calculating the potential energy and the forces on all atoms using
+   the EMT calculator and storing them in an instance of a
+   **CalcResults** data model.
```
francescalb, line 31: Cannot fint this anywhere.

examples/mappings/README.md
```diff
@@ -1,15 +1,40 @@
...
+Simple demos showing how to work with mappings
+==============================================
+This example demonstrates how a user can instantiate a certain data
+model given existing data that is available as instances of other data
+models.
 
+The example is provided at three levels:
 
-This demo demonstrates how a a user can instantiate a certain data model given existing data that is available as instances of other data models.
+1. **simple.py**: Run a simplified example of the example without mapping functions.
```
francescalb, line 9:
```suggestion
1. **simple.py**: Run a simplified example without mapping functions.
                           Note that this does not return an instance of the Molecule (see description of example below),  
                           but rather an instance of CalcResult, which contains both information about the potential energy
                           and forces. 
                          

```

examples/mappings/README.md
```diff
@@ -1,15 +1,40 @@
...
+Example case
+------------
 **Molecule** is the wanted data model.  It is instantiated by:
-1. Loading a given atom structure from a data source into an instance of a **ASEAtoms** data model.
-2. Calculating the potential energy and the forces on all atoms using the EMT calculator and storing them in an instance of a **CalcResults** data model.
-3. Mapping properties of the data models and input/output of mapping functions to ontological concepts.
-4. Invoking `dlite.mappings.instantiate()` to instantiate a **Molecule** data model from the mappings.
+
+1. Loading a given atom structure from a data source into an instance
+   of a **ASEAtoms** data model.
```
francescalb, line 27:
```suggestion
1. Loading a given atom structure from a data source into an instance
   of a **Structure** data model. Note that this is a datmodel describing an
   atomistic structure with symbols and positions of the atoms as well as the unit cell vectors.
```

francescalb: Do also update the Figure to remove references to ASE

examples/mappings/README.md
francescalb, line 31:
```suggestion
2. Load potential energy and the forces as an instance of a
   **CalcResults** data model. 
```

examples/mappings/README.md
```diff
@@ -1,15 +1,40 @@
...
+   **CalcResults** data model.
+
+3. Mapping properties of the data models and input/output of mapping
+   functions to ontological concepts.
+
+4. Invoking `dlite.mappings.instantiate()` to instantiate a
+   **Molecule** data model from the mappings.
 
 The process is shown in the figure below.  Red arrows indicate mappings.
 ![asedemo](asedemo.svg)
```
francescalb, line 40: Change name, there is no ase here.

examples/mappings/requirements.txt
francescalb, line 4:
```suggestion
```

francescalb (changes requested):
Please check the comments.

Most importantly, the description must match the example. There is not ase calculator involved here. 

In simple.py, please remove all namespaces that are not used. They are confusing for a new user. We should only keep code that is used. Also remove the reference to the function_reop in get_instance of simple.py as the function repo is not used. 

In mappingfunc.py we can now import the functions from the mappingfunc_moduel.py in the same way as you to inthe oteexample. Also I suggest saying more explicitly in the comment before ts.add_function that this adds the functions to the ts.fucntion_repo which is explicitly references further down in get_instance.

Finally, the printout from the oteexample prints propene as H6C3 and while mappingfunc give C3H6. I do not see why, right now, but that is a bit confusing.

examples/mappings/README.md
jesper-friis: Updated figure

examples/mappings/mappingfunc_module.py
```diff
@@ -0,0 +1,22 @@
...
+def norm(array, axis=-1):
+    """Returns the norm array along the given axis (default the last)."""
+    # Note that `array` is a Quantity object.  The returned value
+    # will also be a Quantity object with the same unit.  Hence, the
+    # unit is always handled explicitly.  This makes it possible for
+    # conversion function to change unit as well.
+    return np.sqrt(np.sum(array**2, axis=axis))
+
+
+def max(vector):
```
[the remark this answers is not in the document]
jesper-friis: I think `max()` is a very good name of this mapping function. 

There are no risk for confusion with the builtin `max()` function, since they reside in different namespaces:
`buildins.max()` and `mappingfunc_module.max()`.

examples/mappings/requirements.txt
jesper-friis: I see your point. But the purpose of the examples are that you should be able to run the example by just checking it out and install the requirements with pip without needing to compile DLite from source.

Please note that `oteapi-dlite` requires `DLite-Python`, so `DLite-Python` will nevertheless be installed implicitly if we remove it from the requirements file. Better to have it installed explicitly. 

If you want to use the compiled version of DLite, you can do the following:

    pip install -r requirements.txt
    pip uninstall DLite-Python
    cd build-directory
    cmake --install .

jesper-friis: I think it is better to list DLite-Python explicitly rather than having it install implicitly by oteapi-dlite.

examples/mappings/README.md
jesper-friis: done

jesper-friis:
> Please check the comments.
> 
> Most importantly, the description must match the example. There is not ase calculator involved here.
> 
> In simple.py, please remove all namespaces that are not used. They are confusing for a new user. We should only keep code that is used. Also remove the reference to the function_reop in get_instance of simple.py as the function repo is not used.
> 
> In mappingfunc.py we can now import the functions from the mappingfunc_moduel.py in the same way as you to inthe oteexample. Also I suggest saying more explicitly in the comment before ts.add_function that this adds the functions to the ts.fucntion_repo which is explicitly references further down in get_instance.
> 
> Finally, the printout from the oteexample prints propene as H6C3 and while mappingfunc give C3H6. I do not see why, right now, but that is a bit confusing.

Applied all these suggested changes. Some notes:
- After the improved implementation of a function repository as documentation of the `mappingfunc_module`, there is no need to refer to `ts.function_repo` in mappingfunc.py as well.
- The confusing order of propene was due to using set. It is now sorted to always return C3H6.

examples/mappings/mappingfunc_module.py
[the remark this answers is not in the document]
francescalb: But isnæt this bad practice? What if someone wants to use the built in max in the same code further down? This is an example, so mayeb not that important. But at the same time this is where people see how we suggest to use our code. Changing it to maximum will not make it less understandabale and removed the problem.

examples/mappings/requirements.txt
francescalb: OK. But remember that this is an example that other people will use to test things out. If you want to keep it like this, at least add an explanation on what can go wrong in the requirements for this example.

jesper-friis: Added a section to the end of the readme about running the example with DLite installed from source.

examples/mappings/mappingfunc_module.py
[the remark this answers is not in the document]
jesper-friis: Updated requirements.txt such that the example works again.

francescalb (approved):

PR #795, opened 2024-02-21
Corrected syntactic check of id's and uuid's

jesper-friis:
# Description
Corrected syntactic check of id's and uuid's. 

Also rewrote error handling of the return value of  `getuuid()` to correctly handle the new `UUID_EXTRACT` return value. If `getuuid()` can interpret the input UUID, the return value will not be -1 (failure) or 4 (random UUID).

## Type of change
- [x] Bug fix & code cleanup
- [ ] New feature
- [ ] Documentation update
- [x] Test update

## Checklist for the reviewer

### DIFF
--- a/src/dlite-entity.c
+++ b/src/dlite-entity.c
@@ -168,7 +168,7 @@ static DLiteInstance *_instance_store_get(const char *id)
   int uuidver;
   char uuid[DLITE_UUID_LENGTH+1];
   DLiteInstance **instp;
-  if ((uuidver = dlite_get_uuid(uuid, id)) != 0 && uuidver != 5)
+  if ((uuidver = dlite_get_uuid(uuid, id)) < 0 || uuidver == 4)
     return errx(dliteValueError,
                 "id '%s' is neither a valid UUID or a convertable string",
                 id), NULL;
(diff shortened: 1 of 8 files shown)


src/dlite-entity.c
```diff
@@ -168,7 +168,7 @@ static DLiteInstance *_instance_store_get(const char *id)
   int uuidver;
   char uuid[DLITE_UUID_LENGTH+1];
   DLiteInstance **instp;
-  if ((uuidver = dlite_get_uuid(uuid, id)) != 0 && uuidver != 5)
+  if ((uuidver = dlite_get_uuid(uuid, id)) < 0 || uuidver == 4)
```
torhaugl, line 171: I don't fully understand the meaning of 4 here. Is it because UUID_RANDOM = 4? If so, then this should be UUID_RANDOM instead.

torhaugl (commented):
From what I see, there is added new tests and a change in the if statements for uuid version to fix isinstanceuri. Before the tests checked the uuidversion against 5, and now we check against 4. I don't understand why? The tests should instead check against UUID_RANDOM if this is the intended meaning of 4 in various places.

src/dlite-entity.c
jesper-friis: Good point. Fixed

torhaugl (approved):

PR #805, opened 2024-03-14
Added support for onto-ns.com

jesper-friis:
# Description
Added support for fetching instances from http://onto-ns.com/

## Type of change
- [ ] Bug fix & code cleanup
- [x] New feature
- [ ] Documentation update
- [ ] Test update

## Checklist for the reviewer

### DIFF
--- a/.github/workflows/ci_tests.yml
+++ b/.github/workflows/ci_tests.yml
@@ -51,6 +51,9 @@ jobs:
         python3 -m pip install cmake==3.25.2
         cmake --version
 
+    - name: List installed packages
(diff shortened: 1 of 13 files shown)


francescalb (approved):
Nice, but please add

1. a comment about why the save (entities-service) is not tested.
2. an issue about changing the test for fetching ofmr entites-services to fetching  and Instance that has a name that clearly indicates that this is an entity that should not be removed from entities-services. Also, this entity should be somwhere in the repo so taht it canbe easily updalded again to entities-services if it was removed by mistake.

PR #850, opened 2024-06-11
WIP: Do not install optional requirements when building docker images for PyPI

jesper-friis:
# Description
Try to make the building of docker images and PyPI wheels leaner by removing all optional requirements. We only want to build and distribute a DLite Python package that only depends on NumPy. The user can always install the additional requirements later, or by install DLite with 

    pip install dlite-python[full]

The main feature of this PR is the removal of optional Python requirements when building wheels. Other fixes to make the CI tests to pass, includes:
- Fixed building Windows wheel by explicitly specifying Python version setup.py
- Fixed building musllinux wheel by removing support for Python 3.12 on `musllinux_1_2_i686` (a rare platform) since building NumPy fails by complaining about it can't find NumPy. Can be readded in a later PR
- Document our general support Python 3.12 (except for `musllinux_1_2_i686`)
- Made the docker images even leaner by only including a single layer and removing temporary files after build
- Some general code cleanup of setup.py and other files

## Type of change
- [x] Bug fix & code cleanup
- [ ] New feature
- [ ] Documentation update
- [ ] Test update

## Checklist for the reviewer

### DIFF
--- a/.github/docker/Dockerfile-manylinux.template
+++ b/.github/docker/Dockerfile-manylinux.template
@@ -43,7 +43,7 @@ RUN \
     libxslt-devel \
     libxml2-devel && \
   # hdf5-devel is needed for building the hdf5 plugin
-  RUN yum install -y hdf5-devel || true && \
+  yum install -y hdf5-devel || true && \
   # Unpack static libraries
   # It's necessary to be in /opt/_internal because the internal libraries
   # exist here.
--- a/.github/docker/Dockerfile-manylinux.template
+++ b/.github/docker/Dockerfile-manylinux.template
@@ -25,16 +25,14 @@
 # Reference: https://github.com/pypa/manylinux#manylinux2014-centos-7-based
 FROM quay.io/pypa/manylinux{{ TYPE }}_{{ ARCH }}:latest
 
-ARG PY_MINORS="7 8 9 10"
-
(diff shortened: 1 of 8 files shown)


torhaugl (approved):
This PR fixes CI for musllinux and windows. It does not work for musllinux 32bit python 3.12, but that can be fixed in a seperate issue. The fix for Windoes seems to by this FIND_VIRTUALENV CMake variable and specifying the python version. The fix for musllinux was installing openssl-dev.

It also tries to downsize the Dockerfiles by chaining together RUN commands, as well as removing optional requirements during compilation. There is also some cleanup of the code.

It is not fully clear to me that `pip install dlite-python[full]` still works as intended, but if it does, then I think this can be merged. Otherwise, if you are not sure Jesper, I think this can be split into one MR that fixes the CI for musllinux and windows, then another that removes dependencies and reduces the size of the image.

jesper-friis:
> The fix for Windoes seems to by this FIND_VIRTUALENV CMake variable and specifying the python version. The fix for musllinux was installing openssl-dev.

Yes, in addition we had to explicitly specify PYTHON_VERSION in setup.py for Windows.

> It is not fully clear to me that `pip install dlite-python[full]` still works as intended.

Agree, we have to check that. But I think we have to merge and make a new release to test that.

PR #867, opened 2024-06-25
Documented that nothing should be defined outside the plugin class in Python storage plugins

jesper-friis:
# Description
Documented that nothing should be defined outside the plugin class in Python storage plugins,

## Type of change
- [ ] Bug fix & code cleanup
- [ ] New feature
- [x] Documentation update
- [ ] Test update

## Checklist for the reviewer


doc/user_guide/storage_plugins.md
```diff
@@ -175,6 +175,16 @@ See the [Python storage plugin template] for how to write a Python storage plugi
 A complete example can be found in the [Python storage plugin example].
 
 
+:::{danger}
+**When writing a Python storage plugin, do not define any variables or functions outside the `DLiteStorageBase` subclass!**
+
+The reason for this requirement is that all plugins will be loaded into the same shared scope within the built-in interpreter.
+Hence, variables or functions outside the plugin class may interfere with other plugins, resulting in confusing and hard-to-find bugs.
+:::
+
```
francescalb, line 184: Good, but can we add an example in https://github.com/SINTEF/dlite/blob/master/doc/user_guide/storage_plugin.py (doen't matter if it is not used), and in the template?

jesper-friis: > Good, but can we add an example in https://github.com/SINTEF/dlite/blob/master/doc/user_guide/storage_plugin.py (doen't matter if it is not used), and in the template?

Do you mean the note added on line 3 in https://github.com/SINTEF/dlite/blob/81261950afd235dd4c280103bcca798c46ec9463/doc/user_guide/storage_plugin.py?

francescalb: No I was thinking more along the lines 
class plugin....:
   ....
   ...

   def _local_help_function():
       """ An optional class method or a function if desired.
       Note that no functions or variables should be set outside the class
       for a DLite plugin python file.
       ""

jesper-friis: Added an example help method to the storage plugin template

PR #906, opened 2024-08-09
Added test utility functions: importskip(), serverskip()

jesper-friis:
# Description
Added test utility function serverskip() to skip a test if a needed server is down.
Also added importskip() for convenience. 

Motivation for this PR: Avoid that tests fail when services like play.min.io are down.

## Type of change
- [ ] Bug fix & code cleanup
- [ ] New feature
- [ ] Documentation update
- [x] Test update

## Checklist for the reviewer


francescalb (commented):
Looks OK, but I think this will easily end up resulting in tests not running when we should not miss that they are failing. I think this PR might have unintened repercussions.

francescalb (approved):

PR #912, opened 2024-08-11
Fix failing tests on manylinux2014_i686

jesper-friis:
# Description
Fix failing tests on manylinux2014_i686.

Motivation: Fix the segfault we get on github for Python 3.12 on manylinux2014_i686. Unfortunately, I wasn't able to reproduce the segfault when running the docker locally.

While debugging, some issues related compiling on a 32 bit platform were revealed. These are fixed in this PR.

Fixes and changes in this PR:
- Calculate instance hashes consistently across 32 and 64 bit systems. Since this anyway imposed a change in hashes, we used the opportunity to also include dimension and property descriptions in the hash calculations. This is a small change in behaviour. Since transactions haven't been used yet, this change shouldn't affect anyone.
- Allow ctest to use PYTHONPATH to fine external packages that the tests depends on (like tripper, rdflib, ...).
- Allow cmake to procede if hdf5 is not available. hdf5 support will in that case be turned off.
- Compile with treads-support by default (since it is anyway used by Python which is enabled by default).
- Set PATH and PYTHONPATH when running tests for Python scripts.

Unsolved issues:
- Testing of wheels only runs the tests for the Python bindings by running `python bindings/python/tests/test_python_bindings.py`. Ideally we would like to run all tests using `ctest`, but that seems difficult to do... The reason is that `ctest` must be run from the build directory, but the wheel tests are defined by `test-command` in `python/pyproject.toml`. As far as I can see, there is no way to get the build directory from `pyproject.toml`. 

## Type of change
- [x] Bug fix & code cleanup
- [x] New feature
- [ ] Documentation update
- [ ] Test update

## Checklist for the reviewer


francescalb (commented):
Where is the change in the github workflows?

jesper-friis:
> Where is the change in the github workflows?

The failing tests were not run on GitHub CI. The wheel tests only runs a subset of the python tests. This should be fixed when the issue in tripper has been fixed. Should we do that in a new PR?

jesper-friis:
TODO: Run all tests when building wheels, nut just a selection of the python tests

francescalb (approved):
looks OK

PR #939, opened 2024-09-03
Improve installation of Python packages

jesper-friis:
# Description
Improve installation of Python packages and include the following in the distribution:
- utility programs like dlite-getuuid, dlite-codegen, patch-activate.sh
- storages, like blob.json
- README.md and LICENSE files

In addition it fixed a bug in the setup.py script that in addition to dlite-python also installs a pip package called dlite. If you remove that you will remove pip, hense messing with your entire environment.

---

The changes can be tested locally with

    cd python
    python setup.py bdist_wheel

and then install the wheel with pip in a newly created environment 

    pip install dist/DLite_Python-0.5.22-cp310-cp310-linux_x86_64.whl

 After that, you should be able to import dlite in the standard manner

    python
    >>> import dlite

## Type of change
- [x] Bug fix & code cleanup
- [x] New feature
- [ ] Documentation update
- [ ] Test update

## Checklist for the reviewer


bindings/python/tests/global_dlite_state_mod2.py
```diff
@@ -1,4 +1,7 @@
-#!/usr/bin/env python
+# This file is executedevaluated by test_global_dlite_state.py
```
francescalb, line 1:
```suggestion
# This file is executede by test_global_dlite_state.py
```

francescalb (approved):
Just one typo to correct

PR #949, opened 2024-09-12
Improve plugin options

jesper-friis:
# Description
Improve option handling for plugins. This PR tries:
- Remove the limits on possible option values you can pass to storage plugins, due to the restriction imposed by the allowed characters in the query part of an URL. This currently poses a strong limits what data you can pass as an option. 
- Make it more convenient to pass options to plugins (transparent URL encoding).
- Allow option values to be any JSON-serialisable Python object.

Tasks:
- [x] Implement URL-encoding/decoding in C
- [x] Make URL-encoding/decoding accessele to Python
- [x] Let `dlite_option_parse()` transparently URL-decode the option string
- [x] Let `Options` Python class transparently URL-decode the option string
- [x] Add Python API for creating an option string from a dict that transparently URL-encode the values
- [x] Allow to provide options as dict to the high-level Python API, like Storage, and Instance.save(), ...

## Type of change
- [x] Bug fix & code cleanup
- [x] New feature
- [ ] Documentation update
- [ ] Test update

## Checklist for the reviewer


bindings/python/options.py
```diff
@@ -1,44 +1,51 @@
-"""A module for parsing standard options intended to be used by Python
-storage plugins."""
+"""A module for converting option to storage plugins between
+valid percent-encoded URL query strings and and dicts.
```
francescalb, line 2:
```suggestion
"""A module for handling options passed to storage plugins, 
converting between valid percent-encoded URL query strings 
and dicts.
```

bindings/python/options.py
```diff
@@ -1,44 +1,51 @@
...
+    key1=value1;key2=value2...
 
-class Options(dict):
-    """A dict representation of the options string `options`.
+where the keys and and values are percent-encoded.  The key-value
+pairs may be separated by either semicolon (;) or ampersand (&).
+
+Percent-encoding means that all characters that are digits, letters or
+one of "~._-" are encoded as-is, while all other is encoded as their
+unicode byte number in hex with each byte preceeded "%". For example
```
francescalb, line 13:
```suggestion
one of "~._-" are encoded as-is, while all other are encoded as their
unicode byte number in hex with each byte preceeded "%". For example
```

francescalb, line 13:
```suggestion
one of "~._-" are encoded as-is, while all other is encoded as their
unicode byte number in hex with each byte preceeded by "%". For example
```

bindings/python/options.py
```diff
@@ -1,44 +1,51 @@
...
+"a" would be encoded as "a", "+" would be encoded as "%2B" and "å" as
+"%C3%A5".
+
+If a value starts with "%%", the rest of the value is assumed to be a
+percent-encoded json strings.
```
francescalb, line 18:
```suggestion
percent-encoded json string.
```

bindings/python/options.py
```diff
@@ -1,44 +1,51 @@
...
 
-import json
+A URL query string is a string of the form
 
+    key1=value1;key2=value2...
 
-class Options(dict):
-    """A dict representation of the options string `options`.
+where the keys and and values are percent-encoded.  The key-value
+pairs may be separated by either semicolon (;) or ampersand (&).
```
francescalb, line 9: Please give an example.

bindings/python/options.py
```diff
@@ -1,44 +1,51 @@
...
-        opt1=value1;opt2=value2...
 
-    where semicolon (;) may be replaced with an ampersand (&).
+class Options(dict):
+    """A dict representation of the options string `options` with
+    attribute access.
 
-    Default values may be provided via the `defaults` argument.  It
-    should either be a dict or a string of the same form as `options`.
+    Arguments:
```
francescalb, line 31: Arguments? Instance Variables I suppose? And shouldn't this be in the docstring og the __init__?

bindings/python/options.py
jesper-friis: done

bindings/python/options.py
jesper-friis: Yes, you are right, the arguments refer to the arguments passed to the `__init__()` method. But it is common to document arguments in the class docstring. I like to do it for the following reasons:
- In case the class has both a `__new__()` and a `__init__()` method, we don't need to duplicate the argument documentation if we keep it in the class docstring.
- The `__new__()` and `__init__()` methods starts with an underscore and may be hidden by some documentation systems.

bindings/python/options.py
jesper-friis: rewritten

francescalb (approved):

PR #953, opened 2024-09-27
950 load plugins in a separate namespace

jesper-friis:
# Description
Evaluate Python plugins in fully separate namespaces. This will hopefully get rid of the long-standing issue with possible interactions between plugins. This issue became even worse when we started to evaluate the plugins using the calling interpreter.

## Type of change
- [x] Bug fix & code cleanup
- [ ] New feature
- [ ] Documentation update
- [ ] Test update

## Checklist for the reviewer


francescalb (changes requested):
Nice. I only wish for a test. Suggestion: import pyyaml, run dlite with a plugin so that the the yaml plugin is loaded. Do something with pyyaml.

jesper-friis:
> Nice. I only wish for a test. Suggestion: import pyyaml, run dlite with a plugin so that the the yaml plugin is loaded. Do something with pyyaml.

Added test_isolated_plugins.py

francescalb (approved):

PR #985, opened 2024-10-26
Fix json iterate

jesper-friis:
# Description
Cleaned up iteration over instances in a JSON file.

Added `dlite.Instance.get_uuids()` as a convenient way to list instance UUIDs in a storage.

## Type of change
- [x] Bug fix & code cleanup
- [x] New feature
- [ ] Documentation update
- [ ] Test update

## Checklist for the reviewer
This checklist should be used as a help for the reviewer.

- [ ] Is the change limited to one issue?
- [ ] Does this PR close the issue?
- [ ] Is the code easy to read and understand?
- [ ] Do all new feature have an accompanying new test?
- [ ] Has the documentation been updated as necessary?

PR #987, opened 2024-10-27
Avoid overwrite existing YAML files if no change has been made

jesper-friis:
# Description
Avoid overwrite existing YAML files if no change has been made.

The actual fix was to set `self.flushed = True` in the `open()` method of the yaml plugin.

The other changes are general hygienic code cleanup performed while debugging. 

## Type of change
- [x] Bug fix & code cleanup
- [ ] New feature
- [ ] Documentation update
- [ ] Test update

## Checklist for the reviewer


francescalb:
The title of this PR does not correpond to the issue.
The issues is that "DLite overwrites the file if it is not valid". If it is not valid it will make changes to it. 
From the current title it seems that in this case the file with the datamodel will be overwritten.

In case of a valid datamodel, it is not overwritten, in the code as it is now.

jesper-friis:
> The title of this PR does not correpond to the issue. The issues is that "DLite overwrites the file if it is not valid". If it is not valid it will make changes to it. From the current title it seems that in this case the file with the datamodel will be overwritten.
> 
> In case of a valid datamodel, it is not overwritten, in the code as it is now.

You are right, most of the changes are unrelated, but the PR does also fix that yaml files are overwritten. I updated the PR description to reflect that.

francescalb:
When I install it locally it does not solve the issue.

francescalb (approved):
The issue was actually solved in PR #985

francescalb (approved):
Can be merged but fix test

jesper-friis:
> Can be merged but fix test

The test is now fixed.

Issue #978, opened 2024-10-23
dlite-validate output confusing information


francescalb:
dlite-validate returns the full traceback, and thus makes it confusing to understand the validation.

Currently this is the output I get when validating a datamodel with wrong input for dimensions:

> dlite-validate vasp.yaml
Warning: Behavior `singleInterpreter` is not configured. It will be enabled by default from v0.7.0. See https://sintef.github.io/dlite/user_guide/configure_behavior_changes.html for more info.
DLiteOtherError: calling load() in Python plugin 'yaml'
   To see error messages from Python storages, please rerun with the
   DLITE_PYDEBUG environment variable set.
   For example: `export DLITE_PYDEBUG=`

: TypeError: `dimensions` must be either a sequence or a mapping

Only the last line should be shown as the immediate understanding of the above error is that there is an error in the yaml plugin (not that the yaml plugin raises and error).

PR #988, opened 2024-10-27
Address dlite-validate output confusing information

jesper-friis:
# Description
If I understood issue #978 correctly, we should not show the warning about the behavior change from `dlite-validate`. I agree on that.

But how is this solved best? The current implementation expects the user to set environment variable `DLITE_BEHAVIOR=OFF` to turn off all warnings about behavior changes. If we are happy with that, we are done.

If we want tools like `dlite-validate` to never show warnings about behavior changes, we need more. Possible approaches:
- Set `DLITE_BEHAVIOR=OFF` within the tool before importing `dlite`. This is currently implemented in this PR. It is quite a hack and not a very attractive solution...
- If `dlite` is imported from Python, it could look for global variables starting with "DLITE_" and set corresponding environment variables internally. I like this better, since it would be a quite general and nice way to configure `dlite` from Python applications. Now implemented in PR #989 

## Type of change
- [ ] Bug fix & code cleanup
- [x] New feature
- [ ] Documentation update
- [ ] Test update

## Checklist for the reviewer

### DIFF
--- a/bindings/python/scripts/dlite-validate
+++ b/bindings/python/scripts/dlite-validate
@@ -5,7 +5,22 @@ import sys
 import argparse
 import json
 import re
+import ctypes
+from ctypes.util import find_library
 from pathlib import Path
+
+# Silence warning about behavior by setting environment variable
+# DLITE_BEHAVIOR=OFF before importing dlite.
+# This must be done at c-level outside os.environ.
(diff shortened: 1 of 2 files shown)


jesper-friis:
Closing this PR, since a better solution was implemented in PR #989.

PR #989, opened 2024-10-27
Initialise environment from global Python variables

jesper-friis:
# Description
Alternative to PR #988 for fixing issue #978.

This PR allows Python application to configure `dlite` via global variables.

Global variables in (the `__main__` module of) Python that starts with "DLITE_" will be set as environment variables seen by `dlite`. 

Hence, starting your application as

    DLITE_BEHAVIOR=ON python ./app.py

has the same effect as setting `DLITE_BEHAVIOR="ON"` before importing `dlite` in `app.py`.
 
## Type of change
- [ ] Bug fix & code cleanup
- [x] New feature
- [ ] Documentation update
- [ ] Test update

## Checklist for the reviewer


francescalb:
So do I understand correctly that with this change the warnings on bhaviour changes is remvoed by default when using dlite-validate? Can it be added if the users wants to? It would be good to write two sentences about this in the tools documentation (on dlite-validate)

jesper-friis:
> So do I understand correctly that with this change the warnings on bhaviour changes is remvoed by default when using dlite-validate? Can it be added if the users wants to? It would be good to write two sentences about this in the tools documentation (on dlite-validate)

Yes, the behavior warning is now removed in dlite-validate. It is possible to add a separate or extend the --debug option to turn on behavior warnings. In that case we have to process the options before importing dlite. Would you prefer that?

francescalb:
> > So do I understand correctly that with this change the warnings on bhaviour changes is remvoed by default when using dlite-validate? Can it be added if the users wants to? It would be good to write two sentences about this in the tools documentation (on dlite-validate)
> 
> Yes, the behavior warning is now removed in dlite-validate. It is possible to add a separate or extend the --debug option to turn on behavior warnings. In that case we have to process the options before importing dlite. Would you prefer that?

I am not sure we need it. but if it easy do add witht he debug option it might be a good idea.

jesper-friis:
> > > So do I understand correctly that with this change the warnings on bhaviour changes is remvoed by default when using dlite-validate? Can it be added if the users wants to? It would be good to write two sentences about this in the tools documentation (on dlite-validate)
> > 
> > 
> > Yes, the behavior warning is now removed in dlite-validate. It is possible to add a separate or extend the --debug option to turn on behavior warnings. In that case we have to process the options before importing dlite. Would you prefer that?
> 
> I am not sure we need it. but if it easy do add witht he debug option it might be a good idea.

Done. Also made sure that needed cmake configurations are available in dlite-pyembed.c

francescalb (approved):

PR #986, opened 2024-10-26
Validate multiple instances

jesper-friis:
# Description
Update `dlite-validate` to validate sources with multiple instances, without having to provide the `--id` option. 

This PR build on top of PR #985.

**Update**: Added new Python interface to the jstore implemented in C. Right now this increases the code base, but this change makes it possible to e.g. reduce `utils.instance_from_dict()` and `Instance.asdict()` to a few lines of code. An added benefit would be consistency with JSON parsing (since the json parser in C is used under the hood).

Suggests to take the code removal/replacement in a separate PR to have little more control over changes.

## Type of change
- [x] Bug fix & code cleanup
- [x] New feature
- [ ] Documentation update
- [ ] Test update

## Checklist for the reviewer
This checklist should be used as a help for the reviewer.

- [ ] Is the change limited to one issue?
- [ ] Does this PR close the issue?
- [ ] Is the code easy to read and understand?
- [ ] Do all new feature have an accompanying new test?
- [ ] Has the documentation been updated as necessary?

PR #1008, opened 2024-11-20
Validate properties that refer to other datamodels

### LINKED ISSUE #1007: Ref not checked correctly in dlite-validate
When checking a datamodel, whic has a type $ref, it is not actually checked that there is actuallya @ref which has a uri as value. 

In my opinion dlite-validate should check the whole "cascade" of refs as well.

Example of datamodel that is not checked correctly:

uri: http://onto-ns.com/meta/asmod/vasp/0.1/VaspCalculation
description: Documentation of a VASP calculation.
dimensions: {}
properties:
  vaspinput:
    description: Input required for a vasp-calculation
    type: "ref"
    ref: http://onto-ns.com/meta/asmod/vasp/0.1/VaspInputFiles

The last line should be 
$ref: http://onto-ns.com/meta/asmod/vasp/0.1/VaspInputFiles, and dlite-validate should recognize that.

francescalb:
This pull request includes changes to the `dlite-validate` file to improve the validation of datamodels that refer to other datamodels in their properties. 

Question: Should we add error-handling here?

Problem that I do not know how to fix: If there is and error in the dimension in the references datamodel I get a segmentation fault and the whole process closes down without error message.

builds on #986 
closes #1007 

## Type of change
- [x] Bug fix & code cleanup
- [ ] New feature
- [ ] Documentation update
- [ ] Test update

## Checklist for the reviewer

This pull request includes changes to the `bindings/python/scripts/dlite-validate` file to improve the validation process for instances and their properties. The most important changes include adding validation for property references and updating the import and spacing for better readability.


bindings/python/scripts/dlite-validate
```diff
@@ -140,6 +140,22 @@ def main():
     for inst in insts:
         if inst.is_meta:
             check_dimensions(args.url, inst)
+            # Check that datamodels for properties that refer to
+            # a specific datamodel can be loaded.
+            for prop in inst.properties['properties']:
+                if prop.type == 'ref':
+                    if prop.ref == 'ref':
```
jesper-friis, line 147: This is a very specific case to check for. Don't you want to check that `prop.ref` is the URI for a valid datamodel?

Another question, would it make sense to allow `prop.ref` to be empty (just referring to a DLite instance (of any metadata)? There are very few requirements on a generic DLite instance, but one of them is that it must have a `meta` attribute referring to the metadata describing it. It would allow to build more flexible data structures that are not fully defined by the metadata (but who's instances are fully defined).

jesper-friis (commented):
Nice PR. Added some comments

bindings/python/scripts/dlite-validate
```diff
@@ -140,6 +140,22 @@ def main():
...
+                        print(f'Property {prop.name} is a reference, but a '
+                                'uri is not given. Check that the property '
+                                '$ref is given, the $ is needed.')
+                        sys.exit(1)
+                    # Add the path of the original url to storage_path
+                    # since it is likely that referenced data models are 
+                    # in the same folder.
+                    dlite.storage_path.append(Path(args.url).resolve().parent)
```
jesper-friis, line 155: This should work great in the common case that `args.url` is a file path.

However, although not doing any harm, this may not correctly add the storage to `dlite.storage_path` when `args.url` is, say a mongodb storage. In this case, the user will have to use the `--storage-path` option. Maybe the comment (or the 'help' for `--storage-path` option) should be extended with a note about this case?

jesper-friis:
Would be nice with some tests for this PR.

jesper-friis:
> Question: Should we add error-handling here?
> 
> Problem that I do not know how to fix: If there is and error in the dimension in the references datamodel I get a segmentation fault and the whole process closes down without error message.

First, I think that we should never get a segfault from Python. That is a bug that should be fixed independently. If there is an error in the dimensions an exception should be raised.

Regarding your question, dlite should hopefully raise an exception if there is an error. These are already handled, so I don't think we need to do additional error handling in dlite-validate. 

However, if there is an error in the datamodel or instance that is not detected by dlite, we could either handle it here or update the underlying dlite framework to raise a proper exception. I think the choice depends on the nature of the error and its severity.

bindings/python/scripts/dlite-validate
francescalb: This is evry specifc, but it is the case that occurs when the keyu ref is writte as `ref: uri` and not `$ref: uri`. It is a very easy mistake to make, that is why I added it.

francescalb: And yes, I check that the it is valid bu checking that it can be instantiated afterwards.

francescalb:
> If you have time for it, it would be nice with some tests for this PR as well.

In principle I agree, but the setup fpr the placements of tests and how to run them is a bit complicated.

bindings/python/scripts/dlite-validate
francescalb: But in that case you have to give the storage path as argument anyway, no? I updated the doc-string to the --storage-path argument

jesper-friis: I see. Great

jesper-friis:
> First, I think that we should never get a segfault from Python. That is a bug that should be fixed independently. If there is an error in the dimensions an exception should be raised.

PR #1010 should fix the segfault

jesper-friis:
> > If you have time for it, it would be nice with some tests for this PR as well.
> 
> In principle I agree, but the setup and the placements of tests and how to run them is a bit complicated. I have added some new valid and invalid datamodels that can be tested.

See PR #1011 for how to add tests with CMake.

jesper-friis (approved):

PR #1012, opened 2024-11-23
Support for numpy2 

jesper-friis:
# Description
Compile DLite against numpy 2.x. The compiled ABI should be backward compatible with numpy-1.x.

The tests that currently fails seems to be due to that tripper requires pint<0.24, which is not compatible with numpy2.
By force installing pint==0.24.4 the tests succeeds. 

## Type of change
- [ ] Bug fix & code cleanup
- [x] New feature
- [ ] Documentation update
- [ ] Test update

## Checklist for the reviewer


francescalb (approved):
Works forme locally

PR #1013, opened 2024-11-23
Workaround for removed emmo:hasDimension

jesper-friis:
# Description
`hasDimension`, `Dimension` and `hasScalarData` have been removed from EMMO. The two first will be readded after EMMO 1.0.0 has been released. The third must be discussed. 

This PR provides a temporary workaround, keeping the old IRIs while waiting for `hasDimension` and `Dimension` to be readded and a decision for `hasScalarData`.

## Type of change
- [x] Bug fix & code cleanup
- [ ] New feature
- [ ] Documentation update
- [ ] Test update

## Checklist for the reviewer


francescalb:
Why was hasDimension removed from EMMO? If there is a rationale behind that should be reflected in the datadocumentation framework we should think a litlle bit more about this. 
 
 We could also check what the battery ontologies use. I think hasDimension is sued there (at least it was, so maybe they have changed it to something smart)

jesper-friis:
> Why was hasDimension removed from EMMO? If there is a rationale behind that should be reflected in the datadocumentation framework we should think a litlle bit more about this.
> 
> We could also check what the battery ontologies use. I think hasDimension is sued there (at least it was, so maybe they have changed it to something smart)

To readd hasDimension we should create a full conceptualisation of fields. That will not be done before EMMO 1.0.0 has been released. We cannot wait for that, therefore this PR.

PR #1031, opened 2024-12-02
Improved error messages from plugins

jesper-friis:
# Description
Issue a warning message with plugins that fails to load

## Type of change
- [ ] Bug fix & code cleanup
- [x] New feature
- [ ] Documentation update
- [ ] Test update

## Checklist for the reviewer


francescalb:
This is useful, but as you see from the final output message it is not very easy to recognize that the problem can be anything else than a wrong path. In particular it is not clear that if there are errors in the plugin python code it will just fail.  Could we add a sentence to the end of the error message? Suggestion: "If the plugin is listed above, but could not be loaded, an error in the plugin is probable. Please set "export DLITE_PYDEBUG=" and look at the top of the traceback".


image

jesper-friis:
> This is useful, but as you see from the final output message it is not very easy to recognize that the problem can be anything else than a wrong path....

Good point. I have updated the error message now, considering the two cases: the plugin is in the path and plugin is not in the path.

PR #1077, opened 2025-01-22
macOS CMake Updates

nameloCmaS:
# Description

## Library file extensions

MacOS extensions for shared libraries are `.dylib` except for the Python extension loaded by Python (`_dlite.so`) which must follow the Unix naming of `.so` ([not 100% obvious in the docs](https://docs.python.org/3/extending/building.html#building-c-and-c-extensions)).

Changes this one target to be `.so`.

Remove unused variable `pyext_ext` with the target extensions.

Note: This should probably be covered by the `swig_add_library` function call as it deals with the `.pyd` extension on Windows.

## macOS symbols and types

Updated various locations where symbols are defined in alternate headers on macOS.

Changes to types also done as explained in the note in the commit (to do with multiple architectures possible in macOS binaries meaning the type was set as 0 - the note in the section 'Generate integers.h' of `./src/utils/CMakeLists.txt` explains the CMake nuance where multiple architectures are possible). See [docs](https://cmake.org/cmake/help/latest/module/CheckTypeSize.html).

## `LD_LIBRARY_PATH` changes

Changed to add `LD_LIBRARY_PATH` to UNIX only, so `NOT WIN32` becomes `UNIX AND NOT APPLE`.

`LD_LIBRARY_PATH` is not so useful for macOS due to the System Integrity Protect (SIP) which filters this environment variable (except in particular circumstances).

One section in the root `./CMakeLists.txt` file collects these paths and assigns to `PYTHONPATH` on macOS only.

## `dlite_PATH` changes

Changed `dlite_PATH` to `dlite_PATH_NATIVE` in same blocks as it caused a couple of issues.

## configure_file config.h.in

Within `./src/utils/CMakeLists.txt` moved `configure_file(config.h.in config.h)` to the end of the file because it uses variables assigned beneath the original function call position.

The issue was this only worked properly on the second run of the cmake configure.

## Debug parameters for GCC compatible compilers

Added debug parameters for GCC compatible compilers including `Clang` and `AppleClang` and removed case of depreciated `CMAKE_COMPILER_IS_GNUCC`.

## Added `CMAKE_MACOSX_RPATH`

Added `CMAKE_MACOSX_RPATH` as macOS uses shared library search paths slightly differently to Unix.

## Fixed `dlite_PYTHONPATH` string backslash 

This changes the substitution from a string REPLACE (single path) to a list (multiple paths) TRANSFORM.

## `rpl_snprintf` missing on macOS

Changed `rpl_snprintf ` to `snprintf` in `./src/utils/tests/test_snprintf.c`.

`snprintf ` is defined in `compat.h` as

`#define snprintf rpl_snprintf`

where available, and where not available:

`#define snprintf  _snprintf`

## Type of change
- [ ] Bug fix & code cleanup
- [x] New feature
- [ ] Documentation update
- [ ] Test update

## Checklist for the reviewer


jesper-friis:
> Note: This should probably be covered by the `swig_add_library` function call as it deals with the `.pyd` extension on Windows.

Good idea.

We have earlier been hold back on what features in the UseSwig CMake module we could use, because systems with very old versions of swig. However, now when CMake is pip-installable, we can use the latest version of CMake.

src/utils/CMakeLists.txt
```diff
@@ -33,79 +38,79 @@ if(CMAKE_COMPILER_IS_GNUCC)
...
-
-check_symbol_exists(realpath            stdlib.h     HAVE_REALPATH)
-check_symbol_exists(stat                sys/stat.h   HAVE_STAT)
-check_symbol_exists(exec                unistd.h     HAVE_EXEC)
-check_symbol_exists(clock               time.h       HAVE_CLOCK)
-check_symbol_exists(P_tmpdir            stdio.h      HAVE_P_TMPDIR)
-check_symbol_exists(_fileno             stdio.h      HAVE__FILENO)
-check_symbol_exists(isatty              unistd.h     HAVE_ISATTY)
-check_symbol_exists(_isatty             io.h         HAVE__ISATTY)
+check_symbol_exists(strdup              "string.h;_string.h"     HAVE_STRDUP)
```
jesper-friis, line 41: Adding more than one header name seems not to work as expected on Windows. 

The [build wheel test](https://github.com/SINTEF/dlite/actions/runs/12951732700/job/36127426721?pr=1077#step:6:540) on Windows fails, since `HAVE_STRDUP` is not correctly defined, resulting in an redeclaration of `strdup()` in `compat.h`:

```
    D:\a\dlite\dlite\src\utils\compat.h(112,31): warning C4273: 'strdup': inconsistent dll linkage [D:\a\dlite\dlite\python\build\temp.win-amd64-cpython-38\Release\src\utils\dlite-utils.vcxproj]
    C:\Program Files (x86)\Windows Kits\10\Include\10.0.17763.0\ucrt\string.h(536,28): message : see previous definition of 'strdup' [D:\a\dlite\dlite\python\build\temp.win-amd64-cpython-38\Release\src\utils\dlite-utils.vcxproj]
```

jesper-friis:
There is another new issue. `HAVE_PathFileExistsW` is not correctly set on Windows. Results in the following error:

    D:\a\dlite\dlite\src\utils\fileutils.c(708,1): fatal error C1189: #error:  "Windows have no GetFullPathNameW()" [D:\a\dlite\dlite\python\build\temp.win-amd64-cpython-38\Release\src\utils\dlite-utils.vcxproj]

src/utils/CMakeLists.txt
nameloCmaS: No these are not needed, they were added early in the macOS testing when I wasn't sure what the issues with building were. Will remove multiple file paths in each `check_symbol_exists`.

jesper-friis (approved):
Great work!

PR #1078, opened 2025-01-23
Added a possible workaround for the failure in oteapi-dlite

jesper-friis:
# Description
This doesn't look like the right solution. If it solves the problem in oteapi the issue is probably a missing incref() in when the instance is added to the collection.

## Type of change
- [ ] Bug fix & code cleanup
- [ ] New feature
- [ ] Documentation update
- [ ] Test update

## Checklist for the reviewer

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,7 +3,7 @@
 cmake_minimum_required(VERSION 3.16)
 
 project(dlite
-  VERSION 0.5.26
+  VERSION 0.5.27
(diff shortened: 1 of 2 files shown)


jesper-friis:
Closing this PR since it is not the right solution. But it removed the segfault when running pytest in oteapi-dlite.
That the memory test fails just tell us that this is not the right solution, which we already knew. But it supports the hypothesis that there might be a missing incref when the instance is added to the collection...

PR #1192, opened 2025-06-17
Replaced old psycopg2 package with psycopg

jesper-friis:
# Description
The old `psycopg2` package had some issues that could result in segment fault. Replacing it with `psycopg` will hopefully fix that issue.

## Type of change
- [x] Bug fix & code cleanup
- [ ] New feature
- [ ] Documentation update
- [ ] Test update

## Checklist for the reviewer


nameloCmaS:
You need to add

`python-scikit-image:p`

to ci_tests.yml

under the mingw packages.

Package listing on MSYS2 site:

https://packages.msys2.org/packages/mingw-w64-x86_64-python-scikit-image

jesper-friis:
> You need to add
> 
> `python-scikit-image:p`
> 
> to ci_tests.yml
> 
> under the mingw packages.
> 
> Package listing on MSYS2 site:
> 
> https://packages.msys2.org/packages/mingw-w64-x86_64-python-scikit-image

Thank you for the suggestion. However, when trying this I get problems with numexpr. Would it make sense to create a virtual environment for MINGW64 and install all python dependencies there (hoping that there exists pre-build binaries so we don't have to build them every time we tests)?

nameloCmaS:
If that is the case, try adding `python-numexpr:p` also.

However, looking at the last commit action log suggests pyarrow is the issue.

I suggest adding `python-pyarrow:p` also.


Some things I learnt about mingw when I was implementing the workflow action for dlite:

- It could require more packages; you have to keep adding them until all the compiled ones are resolved.
- The mingw maintainers patch their own copies to make them compatible, but you will always be fixed to the current version they have; they don’t have multiple versions available like pip etc do - they are only available from source (but only by changing their git branch).
- You need to install them with pacman first so that pip doesn't need to install them.
- The workflow does use a venv just form mingw I recall - this is because pacman installs its packages in a default python location, but doesn't have python only packages hence the need for pip installs.
- Because you can’t easily fix the version from pacman like you can in pip; the workflow may have issues from time to time, and I assume this will make the maintenance of dlite harder - with possible intermittent failures.

The other option is removing support for mingw; I only added the config because it was partially dealt with in the CMake files already. If sintef don’t have a use case for it, it might be worth removing support.

.github/workflows/ci_tests.yml
nameloCmaS, line 1: Try the following code under line 145:
```
            # First strip out possible packages that could be newer on MinGW or different
            # to the version the constraints set
            sed -E -i 's/^(numpy|pandas|paramiko|pyarrow|pydantic-core|scikit-image|PyYAML)/#\1/g' requirements_full.txt
```

nameloCmaS (changes requested):
Added a potential fix to prevent the conflict.

Seems that paceman on MinGW is installing pyarrow > v21 which is newer than the constrain allows, so pip then goes to install a version within these bounds, but it is not compatible with MinGW without their patching, so the compilation fails.

Fix attempts to comment out packages that need patching but the issue remains that the versions will be outside of the tested parameters.

Suggest that if there is no use case for MinGW in SINTEF that support for it is removed.

Issue #961, opened 2024-10-08
Generate documentation of built-in plugins


jesper-friis:
Add documentation of all build-in plugins.

Ideally this documentation should be generated using the plugin help() function.

**Q**: Is there a need to also generate (parts of) the docstring from option description? 

**A**: Probably not.

Issue #886, opened 2024-07-06
Improve consistency of options for json, yaml and dict representations


jesper-friis:
Overview
- Make options in the yaml plugin consistent with the json plugin
- Make options for asdict() and asjson() consistent with the options for to_bytes() and from_bytes() for the json plugin.

Replace the Python implementation of `dlite.Instance.asdict()` and `dlite.utils.instance_from_dict()` by combining `json.loads()` with `dlite.Instance.to_bytes("json", options=options)` and `json.dumps()` with `dlite.Instance.from_bytes("json", id=id, options=options)`, respectively.

Make the yaml plugin more consistent with the json plugin:
- with_uuid could be renamed to with-uuid for consistency with the json plugin
- it would be more useful to use the single option for controlling for output (this is how it is used in the json plugin)

Avoid as long as possible to break backward compatibility. Old option names should be kept with a deprecation warning.

PR #1305, opened 2026-09-14
Improve plugin system documentation

daniel-sintef:
## Summary

This PR substantially improves the documentation of the DLite plugin system:

- **New: `plugin_system.md`** — an architecture overview of the whole plugin
  system: the three plugin families (storage, protocol, mapping), how they
  compose, the C/Python language layering, plugin lifecycle (lazy
  registration, per-plugin scopes since v0.5.23) and the three discovery
  mechanisms.
- **New: `plugin_reference.md`** — a complete reference of all shipped
  plugins: a table of the 15 storage drivers (language, generic/specific,
  dependencies, key options), per-plugin option tables, and the 4 protocol
  plugins.  Contributes to #961.
- **Rewritten: `mappings.md`** — this page was an empty stub.  It now
  documents both kinds of mappings:
  - *Instance mappings* (mapping plugins): how `Collection.get(label, metaid)`
    and `dlite.mapping()` resolve mappings, cost-based selection, and how to
    write a Python mapping plugin (`DLiteMappingBase` attributes and the
    `map()` method).
  - *Property mappings* (tripper-based): how to add mapping triples and
    mapping functions to a collection triplestore and instantiate new
    instances via `coll.get_instances(..., property_mappings=True)`, plus
    the lower-level `dlite.mappings` functions.
  - A comparison table for choosing between the two.
- **Fixed: `storage_plugins.md`** — the `[protocol plugin]` link pointed to
  the page itself; added cross-links to the new pages.
- **Fixed: `protocol_plugins.md`** — added the missing
  `from dlite.protocol import Protocol` import (the example did not run as
  shown), documented the `dlite.DLiteProtocolBase` base class, the *zip*
  protocol, and corrected the path variable name
  (`dlite.python_protocol_plugin` → `dlite.python_protocol_plugin_path`).
- **Fixed example: `examples/dehydrogenation/python-mapping-plugins/molecule2substance.py`** —
  used the unqualified name `DLiteMappingBase`, which is undefined since
  plugins are evaluated in separate scopes (v0.5.23).  The example failed
  with `DLiteMappingError: cannot map instance labeled 'C2H6' to
  'http://onto-ns.com/meta/0.1/Substance'` and now runs end-to-end.

## Verification

- All code examples in the new pages were executed against the PyPI wheel
  (dlite-python 0.5.49) plus `requirements_full.txt` and
  `requirements_mappings.txt`.
- The fixed dehydrogenation example now runs to completion
  (`Reaction energy: 0.89568 eV`) after running step 1 of the workflow.
- The docs build with Sphinx (user guide section); the new pages render
  correctly, including tables and figures (`storage-protocol.png` is used
  instead of the draw.io SVG, whose `foreignObject` text does not render
  correctly in some viewers).

## Related issues observed while testing (not addressed here)

- `examples/dehydrogenation/3-property-mappings/*/run.py` import
  `make_instance` from `dlite.mappings`, which was removed in commit
  `ce9f5c4c` (mappings cleanup).  These examples are broken on master and
  should be updated to the `instantiate`/`get_instances(property_mappings=True)`
  API.  The new `mappings.md` warns about this.
- Running `examples/dehydrogenation/2-instance-mappings/calculate_reactions.py`
  (after the fix in this PR) completes correctly but segfaults during
  interpreter teardown — appears to be a pre-existing refcount/cleanup issue
  in the C core, unrelated to the documentation changes.
- `pip install dlite-python[full]` warns that the `full` extra does not
  exist on the 0.5.49 wheel; the optional dependencies are listed in
  `requirements_full.txt` instead (see also #1028).  The new
  `plugin_reference.md` documents which plugins are available with which
  installation method.

Contributes to #961.


daniel-sintef:
Dear pi agent the ascii diagram you inserted makes no sense please get rid of it

daniel-sintef:
Removed in commit 849bb60b — the hand-drawn ASCII diagram is gone. The [storage-protocol figure](https://github.com/SINTEF/dlite/blob/master/doc/_static/storage-protocol.png) already shows how the plugin families relate, so the page keeps that one and the bullet-point descriptions.

daniel-sintef:
Honest caveats about this docs PR, for reviewers:

**1. Merge-order dependency of the yaml option names.** bccbbc25 documents the yaml options as `with-uuid`/`with-meta`/`uri-key` — those canonical names are introduced by PR #1306 (issue #886), which is *not merged yet*. On current master, only the underscored spellings are accepted (unknown option keys are silently ignored, so the hyphenated names would be no-ops). If this docs PR merges before #1306, that table is temporarily ahead of the code — the alias note mentions #886, but reviewers may prefer to hold bccbbc25 until #1306 lands, or just merge #1306 first.

**2. Plugin option tables for the less-common drivers are taken from docstrings.** For `minio`, `redis`, `mongodb`, `postgresql`, `csv`, `template` etc., I extracted the option lists from each plugin's `open()` docstring rather than exhaustively re-reading the full implementation of each plugin. The docstrings looked well-maintained, but if a docstring has drifted from the code, the table inherits the drift. The heavily-used ones (`json`, `yaml`, `csv`, `http`) I verified directly against the code.

**3. The "Which kind of mapping to choose" table in mappings.md is my interpretation.** The rows on granularity/units/ontology reflect my reading of the code (`dlite/mappings.py` does handle units via pint, mapping plugins don't), but the "best suited for" advice is editorialising, not documented upstream design intent. Happy to soften or reword if maintainers disagree.

**4. Examples verified, but only against the current release.** All code examples were executed against dlite-python 0.5.49 (PyPI wheel) plus `requirements_full.txt`/`requirements_mappings.txt`. The doctest-style snippets are not wired into any automated doctest runner in this PR — if the maintainers want that, it's probably a separate effort (and `dlite.mapping_plugin_unload('Molecule2Substance')` and friends do run, but only as manual verifications on my side).

PR #1306, opened 2026-09-17
Improve consistency of json/yaml serialisation options

daniel-sintef:
## Summary

First concrete step towards #886 — the foundational fixes that the proposed re-routing of `asdict()`/`asjson()` via `to_bytes("json")` needs to be safe, plus the option-name alignment between the json and yaml plugins asked for in the issue.

### 1. Non-finite floats are now JSON-compatible (and finally round-trippable)

The C serialiser emitted `nan`/`inf` (via printf `%g`), which Python's `json.loads()` rejects — so `json.loads(inst.to_bytes("json"))` failed for any instance containing NaN. Worse, bare `inf` was **rejected by DLite's own JSON parser** (jsmn strict mode only allows primitives starting with `-`, digits, `t`, `f`, `n`), so Infinity values could never be saved to and loaded back from JSON at all.

Now:
- `dlite_type_print()` serialises NaN/+Inf/-Inf as `NaN`/`Infinity`/`-Infinity` — the representations used by Python's json module (and still valid input to `strtod()`/`sscanf()`, so the C scanners read them back).
- The jsmn strict-mode primitive set is extended to accept these tokens.
- Verified: scalar and array properties with NaN/Infinity now save/load round-trip.

### 2. `compact` name clash resolved

In the json plugin, `compact` meant *two different things*: a deprecated alias for `single` in the `open()` options, and "write relations with no newlines" in the `to_bytes()`/memsave options. The latter is now canonically `compact-rel` (matching the existing Python kwarg `compact_rel`), and `compact` is kept there as a deprecated alias that warns.

### 3. `Instance.to_bytes()` accepts option dicts

For symmetry with `from_bytes()`, which already converts dicts via `make_query()`. Previously `to_bytes("json", {"single": "true"})` raised a type error.

### 4. yaml plugin option names aligned with json

`with_uuid`, `with_meta` and `urikey` are renamed to `with-uuid`, `with-meta` and `uri-key`. The old names remain working as deprecated aliases (via `dlite.deprecation_warning()`). Also fixes a copy-paste error in the `to_bytes()` docstring (`single` was documented as "Whether to include UUID in the output").

Aliases are translated **before** applying defaults, so an explicitly-given deprecated option correctly overrides the default value of its canonical name (there is a subtlety here — see the added `_get_options()` helper).

## Testing

- New C test in `test_json.c`: NaN/Infinity/-Infinity serialisation.
- New Python tests in `test_storage.py`: dict options in `to_bytes()`, `json.loads()` compatibility of non-finite floats, and `compact-rel`/`compact` equivalence.
- Full test suite passes locally (109/109, minimal build without HDF5/RDF).

## Not included (needs discussion)

- The polarity renames (`with_parent` → `no-parent`, `soft7` → `arrays`): both invert the meaning, so they change defaults/semantics rather than just names — left for the #886 discussion.
- The actual re-routing of `asdict()`/`instance_from_dict()` through the C json path — this PR deliberately prepares the ground for it (this is why the NaN fix comes first: without it the re-routing would have been broken for non-finite values).
- Note: the DLite instance cache can mask round-trip bugs during testing — `from_url()` returns an already-loaded instance from the instance store without re-reading the file. Worth keeping in mind when verifying the asdict re-routing.

Part of #886.


daniel-sintef:
A few points I'm hesitant about — would appreciate maintainer eyes on these:

**1. The jsmn strict-mode relaxation (src/utils/jsmn.h).** Adding 'N' and 'I' as valid primitive starts slightly weakens JSON strictness: a malformed value like `Nfoo` now tokenizes as a primitive and fails *later* at type conversion ("invalid float: 'Nfoo'") rather than at parse time ("bad token"). I judged this acceptable since every typed conversion rejects garbage anyway, but a maintainer may prefer the parse-time failure. (Python's json module makes the same trade-off when accepting `NaN`/`Infinity` extensions.)

**2. Scope of the `dlite_type_print()` change.** Fixing NaN/Infinity in `dlite_type_print()` changes the output of *every* user of that function, not just the JSON serialiser — e.g. tools/codegen and any table-ish printing. C scanners read `NaN`/`Infinity` back fine (`sscanf`/`strtod` accept them case-insensitively), and Python's `float()` accepts all the spellings too, so I believe this is safe and an overall improvement. But it is a deliberate *global* behaviour change, not a JSON-only one — flagging it in case some downstream consumer prints floats through DLite and parses them with something stricter.

**3. MSVC untested.** The new code uses `isnan()`/`isinf()` type-generic macros and `NAN`/`INFINITY` from `<math.h>`. This is C99, and I'd expect MSVC to handle it, but I have only built/tested on Linux/gcc. The CI Windows build will tell us.

**4. Deprecated-alias precedence in the yaml plugin.** `_get_options()` translates aliases *before* applying defaults, so an explicit `with_meta=true` correctly overrides the default `with-meta=false`. If a user passes *both* `with_meta` and `with-meta` with different values, the canonical name silently wins and the deprecated one is ignored (a warning fires for using it, but not about the conflict). Very much an edge case, but documenting the precedence here.

Things I explicitly verified locally (in case anyone wonders): bool-valued dict options (`{"single": True}`) work through `to_bytes()`; `NaN`/`Infinity`/`-Infinity` round-trip through both the scalar and array property paths, in a *fresh* process (important because the DLite instance cache silently returns already-loaded instances from `from_url()` — this masked the original bug for quite a while during testing!).

storages/python/python-storage-plugins/yaml.py
```diff
@@ -153,19 +196,22 @@ def to_bytes(cls, inst, options=None):
...
+
+        The option names `with_uuid` is a deprecated alias for
+        `with-uuid`.
 
         Returns:
             The bytes (or bytearray) object that the instance is saved to.
         """
-        opts = Options(
-            options, defaults="soft7=true;with_uuid=false;single=true"
+        opts = _get_options(
```
daniel-sintef, line 208: why do you need this strange internal?

